PR #9, opened 2024-03-27
Feature/api creation with FastAPI

hsauve-scottlogic:
- Created FastAPI app
- Created new task for requirements.txt

### DIFF
--- a/backend/api.py
+++ b/backend/api.py
@@ -5,6 +5,7 @@ from fastapi import FastAPI
 app = FastAPI()
 
 
-@app.get("/")
-async def root():
-    return {"message": "Hello World"}
+@app.get("/health")
+async def health_check():
+    return {"message": "Infer GPT is running!"}
+
--- a/backend/api.py
+++ b/backend/api.py
@@ -1,1 +1,10 @@
 # TODO: Add basic FastAPI app for communicating with the front end
+
+from fastapi import FastAPI
+
+app = FastAPI()
+
+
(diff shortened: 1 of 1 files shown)


backend/api.py
```diff
@@ -1 +1,11 @@
 # TODO: Add basic FastAPI app for communicating with the front end
+
+from fastapi import FastAPI
+
+app = FastAPI()
+
+
+@app.get("/health")
+async def health_check():
+    return {"message": "Infer GPT is running!"}
```
harryBedfordSL, line 10: v. minor question, but will pep-8 implement whitespacing rules?

Ch0live: We can configure it not to if it's annoying, but to start I think it will

Ch0live: It's not set up yet however so one to re-look at this when the time comes

bheyman-scottlogic (commented):
Looks good. One thing is the mention in the description of changes to requirements.txt. I cannot see any changes made to this. Was wondering if this should be included in the PR?

backend/api.py
harryBedfordSL: yeah I'm keen for it to enforce consistency when it is set up, cool thanks will approve

harryBedfordSL (approved):

Ch0live:
We've made a new ticket internally for the requirements.txt work

bheyman-scottlogic (approved):

PR #11, opened 2024-03-27
Set Up Mistral LLM

bheyman-scottlogic:
## Description
Sets up Mistral Large to be used in the project with our model deployed in Azure.

## Changelog
- [x] Creates Mistral Large deployment in Azure AI space
- [x] Adds Mistral to project that is connected to our deployed version in Azure
- [x] Adds example `.env` file
- [x] Adds `python-dotenv` to `requirements.txt`


.env.example
```diff
@@ -0,0 +1,2 @@
+AZURE_AI_MISTRAL_LARGE_ENDPOINT=MISTRAL_LARGE_ENDPOINT
+AZURE_AI_MISTRAL_LARGE_KEY=API_KEY
```
Ch0live, line 2: Nit pick but every file should terminate with a new line

Some reasons it's bad are [explained here](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline)

Although I've never encountered an issue myself it's good practice to do this for any file

backend/utils/llm
```diff
@@ -0,0 +1,13 @@
+import os
+from dotenv import load_dotenv
```
Ch0live, line 2: Depending on who merges first this will need adding to the `requirements.txt` file I added. I've added `mistralai` already just not `dotenv`

backend/utils/llm
```diff
@@ -0,0 +1,13 @@
...
+from mistralai.models.chat_completion import ChatMessage
+
+load_dotenv()
+
+endpoint = os.environ["AZURE_AI_MISTRAL_LARGE_ENDPOINT"]
+api_key = os.environ["AZURE_AI_MISTRAL_LARGE_KEY"]
+model = "azureai"
+
+client = MistralClient(api_key=api_key,
+                       endpoint=endpoint)
```
Ch0live, line 13: File is missing `.py` suffix

Ch0live (commented):
Good stuff! I've left a few comments

backend/utils/llm
bheyman-scottlogic: I have updated this, although for some reason in GH it is now showing both files are there

.env.example
```diff
@@ -0,0 +1,2 @@
+AZURE_AI_MISTRAL_LARGE_ENDPOINT=MISTRAL_LARGE_ENDPOINT
```
DannyNicholas, line 1: Is it correct to include `AZURE` in the environment variable name? It's very possible that users will be running this code outside of Azure.

I'm wondering if we namespace env variables to make them specific to our app `INFER_GPT....` to avoid any confusion? Happy to over-ruled if we disagree.

Ch0live (approved):

PR #14, opened 2024-03-28
initialise front end project

harryBedfordSL:
This PR adds a basic frontend for infer-gpt to be able to communicate with the backend server. Note, at this point in time the frontend is not connected to any backend, the bot will just repeat any message sent. Another, smaller PR will be made to do the wiring up.

Overview of code:

- React for UI framework
- Webpack for app bundling
- ESLint for enforcing code consistency
- pure CSS or styling (both modules and global available)

The purpose of this PR is to give us a starting point to work from.

Example screenshot:

image


frontend/.env.example
```diff
@@ -0,0 +1 @@
+BACKEND_ENDPOINT=http://localhost:8080
```
Ch0live, line 1: Dunno if we care but no new line on this & a few other files. Will approve if this is the last comment I don't think it matters too much just raising the spot

bheyman-scottlogic, line 1: For the `.env.example` would this normally be something like =BACKEND_URL ?

frontend/README.md
```diff
@@ -1,6 +1,42 @@
 # Frontend
+UI service for a chat bot.
 
-UI service for InferGPT
+# Run Locally
 
-### To do
-- [ ] Implement initial front end
+Checkout the code, and run the following commands from the root of the frontend directory:
```
Ch0live, line 6:
```suggestion
Checkout the code and run the following commands from the root of the frontend directory:
```

frontend/README.md
```diff
@@ -1,6 +1,42 @@
...
+
+```bash
+npm install
+npm start
+```
+
+The app should now be running on [localhost:3000](http://localhost:8650/).
```
bheyman-scottlogic, line 13: These ports do not match

frontend/src/components/input.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const onSend = useCallback(
+    (event: FormEvent<HTMLElement>) => {
+      event.preventDefault();
+      sendMessage(userInut);
+      setUserInput('');
+    },
+    [sendMessage, userInut]
+  );
+
+  const sendDisabled = useMemo(() => userInut.length === 0, [userInut]);
```
bheyman-scottlogic, line 26: What does useMemo do?

frontend/README.md
```diff
@@ -1,6 +1,42 @@
...
+- create a `settings.json` file in the `.vscode` folder
+- add the following to the `settings.json` file:
+
+```json
+{
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": "explicit"
+  }
+}
+```
```
Ch0live, line 42: Ran `npm run lint` and got this error?

image

[According to this](https://stackoverflow.com/questions/72780296/warning-react-version-not-specified-in-eslint-plugin-react-settings-while-run) it needs this config in `eslintrc.json`?

```
{
  "settings": {
    "react": {
      "version": "detect"
    }
  }
}
```

frontend/webpack.config.js
```diff
@@ -0,0 +1,129 @@
...
+    new MiniCssExtractPlugin(),
+    new webpack.DefinePlugin({
+      'process.env': JSON.stringify(env)
+    })
+  ],
+};
+
+export default config;
+// const HtmlWebpackPlugin = require('html-webpack-plugin');
+// const MiniCssExtractPlugin = require('mini-css-extract-plugin');
```
bheyman-scottlogic, line 69: Is commented out code required here?

bheyman-scottlogic (commented):
A couple of comments/questions, looks great to me though! Well done :)

frontend/src/components/input.tsx
Ch0live, line 26: Is this a typo for `userInput` or something else
```suggestion
  const sendDisabled = useMemo(() => userInput.length === 0, [userInut]);
```

frontend/src/components/message.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  const { class: roleClass, icon } = useMemo(() => roleStyleMap[role], [role]);
+
+  return (
+    <div className={classNames(styles.container, roleClass)}>
+      <img src={icon} />
+      <p>{content}</p>
+    </div>
+  );
+};
```
Ch0live, line 49: So if I understand this correctly, a message to the backend looks like this:

- MessageComponent
   - Takes MessageProps
     - Contains [content, role, time] which are the message content, who said it and the time it was said
   - returns html/JSX which contains the message and roleClass

And the RoleClass converts the Role into the csv and identifier for the bot or the user?

frontend/webpack.config.js
```diff
@@ -0,0 +1,129 @@
...
+    }),
+    new MiniCssExtractPlugin(),
+    new webpack.DefinePlugin({
+      'process.env': JSON.stringify(env)
+    })
+  ],
+};
+
+export default config;
+// const HtmlWebpackPlugin = require('html-webpack-plugin');
```
Ch0live, line 68: Do we need this commented out code?

Ch0live (changes requested):
Followed set up guide in `README.md` and ran perfectly 💯.

Left a few comments but great stuff getting it up and running so fast 🚀

frontend/.env.example
harryBedfordSL: nah I am a fan of new line at end of files, will see if it is easy to add via eslint

frontend/README.md
harryBedfordSL: yup that got rid of the warning, nice one, thanks

frontend/.env.example
harryBedfordSL: I think putting localhost is fine, as it helps suggest what the value would be for local development. I would agree if it was any other sort of sensitive information, like keys or passwords

frontend/README.md
harryBedfordSL: good spot! Will change

frontend/src/components/input.tsx
harryBedfordSL: It is a bit of overkill here, but it is an optimization method that utilizes caching. The variable, `sendDisabled` , will only get recalculated if any of the dependencies change, in this case just `userInput`.

It would be the same to just have `const sendDisabled = userInput.length === 0` , but this will re-run every time the component is re-rendered

frontend/webpack.config.js
harryBedfordSL: no, good point! That is me not reviewing my own stuff well. Is not meant to be there, will remove

frontend/src/components/input.tsx
harryBedfordSL: is a typo, thanks

frontend/src/components/message.tsx
harryBedfordSL: not quite, the `roleClass` is just a CSS class that determines the position of the message in the chat. 

Currently, `server.ts` is the only file that will interact with the backend, and atm is just taking a `message: string` as the only argument. We can change this to whatever is needed in the wiring up step

frontend/webpack.config.js
harryBedfordSL: Ben spotted too, am removing

frontend/.env.example
harryBedfordSL: okay I've added a eslint plugin that does this for all of our js, ts, tsx files, and have manually done the rest. Can switch out plugins later if we want something else

frontend/src/components/message.tsx
Ch0live: Cool cool, in that case for my first connection PR I'll just adjust the `server.ts` to take the response I sent it from the backend 👍

Ch0live (approved):

bheyman-scottlogic (approved):

PR #15, opened 2024-03-28
Connect ChatBot to backend API

Ch0live:
## Description

Connects the endpoint for the API to methods that can call a deployed Mistral-Large instance in Azure and return a generated response.

image

## Changelog
- Re-prioritize headings in `README.md`
- Add `node_modules` directory to `.gitignore`
- New `/chat` endpoint that returns LLM response. ChatBot response are aware the ChatBot is called InferGPT and some of it's capabilities


backend/utils/llm.py
```diff
@@ -1,12 +1,32 @@
...
 
 load_dotenv()
 
-endpoint = os.environ["INFER_GPT_MISTRAL_LARGE_ENDPOINT"]
-api_key = os.environ["INFER_GPT_MISTRAL_LARGE_KEY"]
-model = "azureai"
+try:
+  endpoint = os.environ["INFER_GPT_MISTRAL_LARGE_ENDPOINT"]
+  api_key = os.environ["INFER_GPT_MISTRAL_LARGE_KEY"]
+  key = os.getenv("OPENAI_API_KEY")
```
hsauve-scottlogic, line 11: Are we also using OpenAI?

backend/director.py
```diff
@@ -1,3 +1,18 @@
...
+from utils.llm import call_model
+
+prompt= """
+You are a Chat Bot called InferGPT.
+
+You're abilities include:
+
+- Setting and tracking Goals
+- Remembering conversation information from previous sessions and using it in future discussions
+
```
hsauve-scottlogic, line 10: Would it be useful to add something like ‘if the user has not provided a goal, do not make up one’, and also provide it with a few examples of what a goal is, or is that overengineering at this point?

backend/utils/llm.py
```diff
@@ -1,12 +1,32 @@
...
+      model=model,
+      messages=[
+         ChatMessage(
+            role="user",
+            content=prompt
+         )
+      ],
+      max_tokens=125
+   )
+   print(response.choices[0].message.content)
```
harryBedfordSL, line 31: Is this a  dev comment or intended logging?

hsauve-scottlogic (commented):
Looks great Charlie, just left a couple of comments

backend/utils/llm.py
```diff
@@ -1,12 +1,32 @@
...
 load_dotenv()
 
-endpoint = os.environ["INFER_GPT_MISTRAL_LARGE_ENDPOINT"]
-api_key = os.environ["INFER_GPT_MISTRAL_LARGE_KEY"]
-model = "azureai"
+try:
+  endpoint = os.environ["INFER_GPT_MISTRAL_LARGE_ENDPOINT"]
+  api_key = os.environ["INFER_GPT_MISTRAL_LARGE_KEY"]
+  key = os.getenv("OPENAI_API_KEY")
+  model = "azureai" # might me mistral-large
```
bheyman-scottlogic, line 12: In our case this is `azureai`, but as Danny mentioned on a previous PR, this may be different. Maybe should be changed to an environment variable.

backend/utils/llm.py
bheyman-scottlogic, line 11: Use both `os.getenv` and `os.environ`. Should stick to one for consistency. Don't believe it makes a difference which one.

backend/utils/llm.py
bheyman-scottlogic, line 31: Should this print statement be removed?

backend/director.py
```diff
@@ -1,3 +1,18 @@
...
+Introduce yourself if asked.
+"""
```
bheyman-scottlogic, line 12: Does this prompt cover too much at this point as some of that functionality is not included? Also can set this as a system prompt instead of adding to the user prompt

```python
messages=[
            {"role": "system", "content": system_message},
            {"role": "user", "content": user_prompt}
        ],
```

bheyman-scottlogic (commented):
Looks good Charlie 💯  Was great to see it up and running with you as well. Just a couple of things to look at from me!

backend/utils/llm.py
Ch0live: Not yet, removed

backend/utils/llm.py
Ch0live: It is! Removed

backend/director.py
Ch0live: I think we do that when this fails to set the goal at a later date. We don't want to overuse tokens if we don't need to

backend/utils/llm.py
Ch0live: Changed

backend/utils/llm.py
Ch0live: Removed

backend/director.py
Ch0live: Adjusted to separate system and user prompts

hsauve-scottlogic (approved):

bheyman-scottlogic (approved):

PR #17, opened 2024-04-02
Wire up InferGPT frontend and backend

Ch0live:
## Description
Connects the backend /chat and /health endpoints to the frontend. The /health endpoint can be accessed by sending the message "healthcheck" to InferGPT. You will either receive a "InferGPT backend is healthy!" or "Error: InferGPT backend is unhealthy" in response.

healthcheck_keyword_demo

All other messages go to the LLM for a generated response

hello_infergpt

## Changelog
- Changed `server.ts` to call the backend on port `8250`
- Added a check for if the user submitted the string "healthcheck". If so, the `/health` endpoint is called
- Allowed CORS requests (as both services are currently running on the same machine)
- Tweaked the system prompt to avoid the service continually regurgitating it's initial hello statement

### DIFF
--- a/backend/api.py
+++ b/backend/api.py
@@ -1,8 +1,21 @@
 from fastapi import FastAPI
 from fastapi.responses import JSONResponse
 from director import question
+from fastapi.middleware.cors import CORSMiddleware
 
 app = FastAPI()
+
+origins = [
+    "http://localhost:8650",
+]
+
+app.add_middleware(
+    CORSMiddleware,
+    allow_origins=origins,
+    allow_credentials=True,
+    allow_methods=["*"],
+    allow_headers=["*"],
+)
 
 @app.get("/health")
 async def health_check():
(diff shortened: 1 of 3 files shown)


DannyNicholas:
Looks good to me! 👍 

image

backend/api.py
```diff
@@ -1,9 +1,22 @@
 from fastapi import FastAPI
 from fastapi.responses import JSONResponse
 from director import question
+from fastapi.middleware.cors import CORSMiddleware
 
 app = FastAPI()
 
+origins = [
+    "http://localhost:8650",
```
DannyNicholas, line 9: Fine for now but ideally we'd want this to be configurable.

DannyNicholas (approved):

backend/api.py
Ch0live: Will amend on a new PR

bheyman-scottlogic (approved):
Happy with everything there, good work!

PR #18, opened 2024-04-02
Set Up Graph Database

bheyman-scottlogic:
## Description
Database is set up and the application can now write to the Neo4j database

## Changelog
- Update `README.md` to include instructions for setting up Neo4j Desktop
- Add schema to `README.md`
-  Add `graph_db_utils.py` with connection to db and function to create new node


data/graph_db_utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+driver = GraphDatabase.driver(URI, auth=AUTH)
+
+def test_connection():
+    try:
+        driver.verify_connectivity()
+        print("database connection active")
+        
+    except Exception as e:
+        print(f"Error: {e}")
```
DannyNicholas, line 18: It looks like when we get an exception, it will just print a message and then carry on. The calling code could think the request has succeeded. I suggest we'd want to throw an exception and stop execution in this case (using `raise` I think).

Same suggestion for other exception block

data/graph_db_utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+URI = os.getenv("NEO4J_URI")
+AUTH = (os.getenv("NEO4J_USERNAME"), os.getenv("NEO4J_PASSWORD"))
+
+driver = GraphDatabase.driver(URI, auth=AUTH)
+
+def test_connection():
+    try:
+        driver.verify_connectivity()
+        print("database connection active")
```
DannyNicholas, line 15: Can we replace all `print` statements with logging methods. We can use `debug` level if they are lower priority messages that we may want to filter out when reviewing log messages.

data/graph_db_utils.py
```diff
@@ -0,0 +1,44 @@
...
+    finally:
+        driver.close()
+        
+
+
+def create_node(name, description):
+    try:
+        session = driver.session()
+        query = """
+        CREATE (g:Goal {name: $name, description: $description})
```
DannyNicholas, line 29: I've read that `MERGE` will create nodes if they don't exist or take no action if the node already exists.

I'm not 100% sure on what best practice is but wondering if we want to use `MERGE` to avoid potentially creating duplicate nodes?

data/graph_db_utils.py
```diff
@@ -0,0 +1,44 @@
+import os
```
Ch0live, line 1: Did we agree at stand up today that we would move this into the /backend space?

data/README.md
```diff
@@ -2,9 +2,34 @@
...
+
+## Using Neo4j Desktop
+
+If using Neo4j Desktop, this can be downloaded [here](https://neo4j.com/download/)
+
+Once installed follow the steps below:
+
+1. Create new project to host your DBMS.
+2. Create a new database within the project.
+3. Add the username and password to the `.env` file.
```
Ch0live, line 15: I think it should be specified that you need to create another `.env` file here - might be just me adding the config settings the `backend/.env` but this tripped me up

Ch0live: Unless I'm missing something? Just realised the `.example.env` in the backend is the one with the example property values so is the config intended to live there?

data/graph_db_utils.py
```diff
@@ -0,0 +1,44 @@
...
+def test_connection():
+    try:
+        driver.verify_connectivity()
+        print("database connection active")
+        
+    except Exception as e:
+        print(f"Error: {e}")
+        
+    finally:
+        driver.close()
```
Ch0live, line 21: Good use of `finally:` 🚀

Ch0live (changes requested):
Nice to the point `graph_db_utils.py`. A few comments but nothing major 👍

data/README.md
bheyman-scottlogic: Moved the `graph db utils` to the `backend/utils/` as this seems to make more sense

data/graph_db_utils.py
```diff
@@ -0,0 +1,44 @@
...
+        
+
+
+def create_node(name, description):
```
philbaxter: Would this be better named `create_goal`?

philbaxter (changes requested):

Ch0live (approved):

DannyNicholas (approved):

PR #19, opened 2024-04-03
Add Config class

Ch0live:
## Description
Off the back of [Danny's comment from last PR](https://github.com/WaitThatShouldntWork/InferGPT/pull/17#discussion_r1547852632) this change adds a class to load in all the config from the .env files and makes it easier and simpler to import said config around the backend.

## Changelog
- Add `config.py` with class Config to load and serve configuration around InferGPT backend
- Adjust imports to use the `__init__` file within /utils (see using packages [here](https://docs.python.org/3/tutorial/modules.html#packages
- Fix a typo in `useMessage.ts`
- Renamed a few variables mostly to shorten them


backend/api.py
```diff
@@ -1,12 +1,14 @@
...
+from utils import Config
 from director import question
 from fastapi.middleware.cors import CORSMiddleware
 
 app = FastAPI()
+config = Config()
 
 origins = [
-    "http://localhost:8650",
+    "http://localhost:" + config.port,
```
DannyNicholas, line 11: Should the host also be part of the configuration. We're only running the UI locally for now but eventually the UI may  be hosted elsewhere (such as Azure).

So we'd have....
```
origins = [
    config.host + ":" + config.port,
]
```

backend/.env.example
```diff
@@ -1,3 +1,7 @@
+# Required properties
 INFER_GPT_MODEL="azureai"
-INFER_GPT_MISTRAL_LARGE_ENDPOINT=MISTRAL_LARGE_ENDPOINT
+INFER_GPT_MISTRAL_LARGE_URL=MISTRAL_LARGE_ENDPOINT
 INFER_GPT_MISTRAL_LARGE_KEY=MISTRAL_LARGE_API_KEY
+
+# Optional properties (with default values)
+# INFER_GPT_HOST="8650"
```
DannyNicholas, line 7: Strictly speaking this the port (not the host). I think we'd want both in the configuration anyway...

Could I also suggest that we don't comment them out to prove they work successfully.

```
# Optional properties (with default values)
INFER_GPT_HOST="http://localhost"
INFER_GPT_PORT="8650"
```

DannyNicholas (commented):
This is great work @Ch0live - big fan of separating out configuration into it's own class. 

A couple of minor comments.

backend/api.py
DannyNicholas: Alternatively, we stick to one `config.host` variable and that holds the entire URL including the port?

Ch0live: I've updated the code to have a `config.frontend_url` instead and also made some of the names more consistent

backend/.env.example
Ch0live: This has been renamed as `INFER_GPT_FRONTEND_URL` and fixed

bheyman-scottlogic (approved):
Looks good, nice work!

PR #22, opened 2024-04-08
Log Exceptions

Ch0live:
## Description
We previously could not see exceptions in the logs of the backend. This was because exceptions were caught in the controller and were not logged manually after being caught. This has been amended in this PR. Here's an example where I pasted in a `print(abc)` inside the `llm.py` without defining `abc`

image

## Changelog
- If an error occurs, the back end will log the exception and stacktrace
- Improves front end messages


philbaxter (approved):

frontend/src/server.ts
```diff
@@ -3,11 +3,15 @@ export interface ChatMessageResponse {
...
+  message: 'Healthcheck: InferGPT backend is healthy!'
 };
 
 const unhappyResponse: ChatMessageResponse = {
-  message: 'Error: InferGPT backend is unhealthy'
+  message: 'Healthcheck: InferGPT backend is unhealthy'
+};
+
+const unhappyChatResponse: ChatMessageResponse = {
+  message: 'I\'m sorry, but I was unable to get a response from InferGPT. Please check the status of the service using the phrase \"healthcheck\"'
```
bheyman-scottlogic, line 14: Bit pinickity but saying "I was unable to get a response from InferGPT" implies separate systems? Isnt the whole thing called InferGPT?

bheyman-scottlogic (commented):
One small comment, but overall looks good to me!

bheyman-scottlogic (approved):

philbaxter (approved):

PR #23, opened 2024-04-08
Create Goal Agent

bheyman-scottlogic:
## Description
When the method is called in the goal agent, the goal and description will be extracted from the user input and persisted to the graph database.

## Changelog
- Goal agent created
- Function to extract goal and description added
- Goal agent linked to the db function to persist goals to db
- Update config and .env naming for llm model
- Added logging functionality


backend/utils/llm.py
```diff
@@ -6,12 +7,12 @@
 logger = logging.getLogger(__name__)
 config = Config()
 
-client = MistralClient(api_key=config.mlarge_key, endpoint=config.mlarge_url)
```
harryBedfordSL, line 9: Forgive me if I'm missing something, but why do we no longer need the endpoint argument?

backend/utils/llm.py
```diff
@@ -24,5 +25,28 @@ def call_model(system_prompt, user_prompt):
...
+         ),
+         ChatMessage(
+            role="user",
+            content=user_prompt
+         )
+      ], 
+      tools=tools, 
+      tool_choice="any"
+   )
+   logger.info("{0} response : \"{1}\"".format(config.mistral_model, response.choices[0].message.content))
```
harryBedfordSL, line 48: a lot of this is repeated. I wonder if it would be neater to split out into several methods. I'm thinking `getResponse` , `call_model` and `call_model_with_tools` and then we'd only need to expose `call_model` and `call_model_with_tools` . (this would only work is tools can be defaulted to `None` in the `getResponse` method)

backend/agents/goal_agent.py
```diff
@@ -0,0 +1,52 @@
...
+                    },
+                },
+                "required": ["goal_name"],
+            },
+        },
+    }
+]
+ 
+names_to_functions = {
+    'create_goal': functools.partial(graph_db_utils.create_goal),
```
harryBedfordSL, line 35: this looks cool, what is `functools.partial` doing here?

backend/utils/llm.py
bheyman-scottlogic: As we are using the normal mistral api for this now opposed to the azure deployment. This is due to the function calling features that are not available in the azure space at this time. It therefore doesn't require the endpoint

backend/agents/goal_agent.py
bheyman-scottlogic: Where the string `create_goal` appears, it assigns the function as the dictionary value. The function can then be called. You can see this being called on line 52

backend/utils/graph_db_utils.py
```diff
@@ -23,14 +21,14 @@ def test_connection():
...
+def create_goal(goal_name, goal_description):
     try:
         session = driver.session()
         query = """
-        MERGE (g:Goal {name: $name, description: $description})
+        MERGE (g:Goal {goal_name: $goal_name, goal_description: $goal_description})
         RETURN g
         """
-        result = session.run(query, name=name, description=description)
+        result = session.run(query, goal_name=goal_name, goal_description=goal_description)
```
Ch0live: I haven't tried it so correct me if I'm wrong but you may be able to pass the arguments without the `=` operator as they have the same name as the parameters
```suggestion
        result = session.run(query, goal_name, goal_description)
```

Ch0live (changes requested):
Nice stuff creating our first agent 🚀. Left a few comments

hsauve-scottlogic (commented):
Great work Ben :)

backend/utils/graph_db_utils.py
bheyman-scottlogic: Does not work, also discussed on call

philbaxter (approved):

harryBedfordSL (dismissed):
love it, thanks for the changes

harryBedfordSL (approved):

Ch0live (approved):

PR #24, opened 2024-04-09
Add frontend linting to PR pipeline

harryBedfordSL:
This PR:

- adds linting for our frontend code to the PR pipeline (we should then make it a mandatory step to pass before merging, and should do the same for the backend)
- does the small lint changes needed for the frontend
- changes the expected line endings to unix for the front end, to align with ci environment

### DIFF
--- a/.github/workflows/lint-frontend.yml
+++ b/.github/workflows/lint-frontend.yml
@@ -17,10 +17,10 @@ jobs:
           node-version: '14'
 
       - name: Move to frontend directory
-        run: 
-          - ls
-          - cd frontend
-          - ls
+        run: |
+          ls
+          cd frontend
+          ls
 
       - name: Install dependencies
         run: npm install
--- a/.github/workflows/lint-frontend.yml
+++ b/.github/workflows/lint-frontend.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Move to frontend directory
         run: |
           ls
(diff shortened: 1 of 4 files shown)


frontend/.eslintrc.json
```diff
@@ -1,41 +1,41 @@
...
+  ],
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "ecmaVersion": "latest",
+    "sourceType": "module"
+  },
+  "plugins": ["@typescript-eslint", "react", "@stylistic/js"],
+  "rules": {
+    "indent": ["error", 2],
+    "linebreak-style": ["error", "unix"],
```
harryBedfordSL, line 31: this is the only change in this file. The github action that runs the lint is in a unix system, so we should do the same for local development (this means using LF as the end of line sequence)

.github/workflows/lint-frontend.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: 21
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Lint
+        run: npm run lint
```
Ch0live, line 26: Neat! 🚀

Ch0live (approved):

PR #25, opened 2024-04-10
Connect director to goal agent

bheyman-scottlogic:
## Description
Goal agent is now connected to the director and the full workflow from frontend to database is complete.

## Changelog
- Director will now decide whether user question is a goal
- Added in additional prompts

### DIFF
--- a/backend/director.py
+++ b/backend/director.py
@@ -1,14 +1,35 @@
 import logging
 from utils.llm import call_model
+from agents.goal_agent import create_user_goal
 
 logger = logging.getLogger(__name__)
 
 system_prompt= """
 You are a Chat Bot called InferGPT.
-You're abilities include:
-- Setting and tracking Goals
+Your sole purpose is to get the user to tell you their goal.
+If the user does not provide a goal, ask them to provide a goal 
+"""
+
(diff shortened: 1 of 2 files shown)


Ch0live (dismissed):

backend/director.py
```diff
@@ -1,14 +1,35 @@
...
 
 def question(question):
   logger.info("director calling call_model method")
-  return call_model(system_prompt, user_prompt=question)
+  response = determine_intent(question)
+  if response == "FALSE":
+    return call_model(system_prompt, user_prompt=question)
+  else: 
+    goal_saved = create_user_goal(question)
+    return "I have created a goal for you. Goal: {0}. Description: {1}".format(goal_saved['name'], goal_saved['description'])
```
harryBedfordSL, line 32: I feel like instead of response this variable name should indicate that it is like a boolean e.g. `hasIntent`. I wonder whether it would work to convert it into a boolean in one line too i.e. 

```python
hasIntent = determine_intent(question) == 'TRUE'
```

Also, the `else` is not needed because of the return in the `if`.

Ch0live (approved):

harryBedfordSL (approved):

PR #27, opened 2024-04-12
Feature/backend linting

bheyman-scottlogic:
## Description
Add backend linting

## Changelog
- Add GitHub action for backend linting
- Update linting errors
- Add readme

### DIFF
--- a/.github/workflows/lint-backend.yml
+++ b/.github/workflows/lint-backend.yml
@@ -1,0 +1,22 @@
+name: Lint Backend
+on: 
+  pull_request:
+
+jobs:
+  linting:
+    name: Linting Backend
+    defaults:
+      run:
+        working-directory: ./backend
+    runs-on: ubuntu-latest 
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+
+      - name: Lint and Fix
+        uses: chartboost/ruff-action@v1
+        with:
+          args: --check .
--- a/backend/README.md
+++ b/backend/README.md
@@ -12,3 +12,29 @@ Python service for the InferGPT backend.
 - To install all project dependencies (listed in `requirements.txt`) run `pip install -r requirements.txt`.
 - Once all dependencies have been pulled run `uvicorn api:app --port 8250` to start the app. 
 - Check the backend app is running at [http://127.0.0.1:8250/health](http://127.0.0.1:8250/health).
+
+## Backend Linting
+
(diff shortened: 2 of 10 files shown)


.github/workflows/lint-backend.yml
```diff
@@ -0,0 +1,19 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+
+      - name: Lint
+        uses: chartboost/ruff-action@v1
+        with:
+          src: ./backend
```
harryBedfordSL, line 19: nice!

backend/README.md
```diff
@@ -12,3 +12,35 @@ Python service for the InferGPT backend.
...
+If it is not installed use the follwoing command:
+
+```bash
+pip install ruff
+```
+
+If ruff is installed correctly, the python files in the backend will be checked using the following command
+
+```bash
+ruff check backend
```
harryBedfordSL, line 37: is `backend` here needed if you are already in the backend directory?

backend/utils/config.py
```diff
@@ -1,12 +1,13 @@
...
+
 class Config(object):
     def __init__(self):
         self.frontend_url = None
-        self.mistral_url = None
-        self.mistral_key = None
-        self.mistral_model = None
+        self.mistralruf_url = None
+        self.mistralruf_key = None
+        self.mistralruf_model = None
```
harryBedfordSL, line 10: why are we adding `ruf` to the property names here?

backend/README.md
bheyman-scottlogic: No, should I assume that the person is in the backend directory then?

backend/utils/config.py
bheyman-scottlogic: Yeah I will change this, think it might have done this on one of the auto-fixes, but these have now been removed as well

backend/README.md
harryBedfordSL: I think from the backend readme, yes :slightly_smiling_face: but can always add before the code block from this directory run:   sort of thing

harryBedfordSL (approved):

hsauve-scottlogic (approved):

.github/workflows/lint-backend.yml
```diff
@@ -0,0 +1,19 @@
+name: Lint Backend
+on: 
+  pull_request:
```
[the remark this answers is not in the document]
bheyman-scottlogic: I believe it is ok without doing this as will just perform on changes to a pr or when its uploaded by default which is what we want

PR #28, opened 2024-04-15
Dockerize app

Ch0live:
## Description

Adds a `compose.yml` that can orchestrate the creation and running of 2 Docker images. *Needs to be verified with a valid API key*

## Changelog
- Change `INFER_GPT_BACKEND_URL` to `BACKEND_URL`
- Adds docker files to enable `docker-compose up` as a single line command to start the app
- Fix typo in `.example.env`
- Add better healthcheck logic with improved logging and messages on the frontend
- Refactor `server.ts`


backend/.env.example
```diff
@@ -2,8 +2,8 @@
 MODEL="azureai"
 MISTRAL_LARGE_URL=MISTRAL_LARGE_ENDPOINT
 MISTRAL_LARGE_KEY=MISTRAL_LARGE_API_KEY
-NEO4j_URI="bolt://localhost:7687"
```
DannyNicholas, line 5: `localhost` won't work when the backend is running in a Docker container but neo4j is running outside Docker.

To make a request from a container to the local host, we must use `host.docker.internal`. 

We need:
```
NEO4J_URI="bolt://host.docker.internal:7687"
```

backend/.env.example
```diff
@@ -2,8 +2,8 @@
 MODEL="azureai"
 MISTRAL_LARGE_URL=MISTRAL_LARGE_ENDPOINT
 MISTRAL_LARGE_KEY=MISTRAL_LARGE_API_KEY
```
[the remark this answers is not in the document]
Ch0live: > Calls to Mistral fail as these environment variables don't match the names used in config.py.

I have updated the config so the .env.example now matches this

> At the time of writing llm.py ignores the Mistral URL as it defaults to the Mistral cloud instance but we should leave it in to support Azure hosted Mistrals in future.

This is a good point, we should make the `MODEL` parameter swap between calling the API and calling an Azure instance. I'll add this as a small ticket to our backlog

backend/.env.example
DannyNicholas: @Ch0live - how to you want to handle this?

I think we either change `NEO4J_URI` in the `.env.example` or we include a comment in the `README` on how to configure the `NEO4J_URI` when the backend is running within Docker (and neo4j is not).

Ch0live: Still getting an error
```
Error: {code: Neo.ClientError.Security.Unauthorized} {message: The client is unauthorized due to authentication failure.}
```
Unsure on the fix for this :/

Ch0live:
Added better healthcheck logic to make checking your set up when running with docker easier

image

And verified with a correct set up that all works as intended

image

backend/.env.example
Ch0live: Fixed! It was a permission issue due to an incorrect password

DannyNicholas (approved):

DannyNicholas:
@Ch0live - when you merge can you also rebase the `release` branch on top of these changes so we also get the "dockerisation" benefits in our release branch?

Ch0live:
Yes good point will do 👍

PR #29, opened 2024-04-17
Add backend test framework

harryBedfordSL:
## Description

This PR adds `pytest` and `pytest-mock` to our backend, allowing us to write unit tests

## Changelog
-  add `pytest` and `pytest-mock` to requirements file
- add basic tests for `api.py`
- add basic tests for `llm.py`
- update readme on how to run tests

### DIFF
--- a/backend/api_test.py
+++ b/backend/api_test.py
@@ -4,15 +4,17 @@ from api import app, healthy_response, error_message
 client = TestClient(app)
 utterance = "Hello there"
 
+
 def test_health_check_response():
     response = client.get("/health")
 
     assert response.status_code == 200
     assert response.json() == healthy_response
 
+
 def test_chat_success_response(mocker):
-    expected_message = 'InferGPT backend is healthy'
-    mock_question = mocker.patch('api.question', return_value=expected_message)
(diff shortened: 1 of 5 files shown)


backend/api_test.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def test_health_check_response():
+    response = client.get("/health")
+
+    assert response.status_code == 200
+    assert response.json() == healthy_response
+
+
+def test_chat_success_response(mocker):
```
DannyNicholas, line 15: How is this `mocker` injected into the tests?

DannyNicholas (approved):

hsauve-scottlogic (approved):
Looks very good Harry, nice to see how Pytest-mock works!

backend/api_test.py
harryBedfordSL: so my understanding is that with `pytest` you can make [fixtures](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html) that are always injected into each test. `pytest-mock` is a plugin that provides the `mocker` as a fixture

PR #30, opened 2024-04-17
happy path mermaid sequence diagrams

Ch0live:
## Description

Adds diagrams for the happy paths in the next sum my subscriptions epic

## Changelog
- Adds new Markdown file with mermaid diagrams listed inside


Ch0live:
I was curious if we should 
- add a final summariser method at the top level
- include the scratchpad here when saving between runs? Might be useful for the summariser I mentioned above

assets/sum-my-subscriptions/happy_flows.md
```diff
@@ -0,0 +1,83 @@
...
+## "Hello! What can you do?"
+
+I am a user who is asking about the bots capabilities
+
+### Find tasks from question
+```mermaid
+sequenceDiagram
+    Frontend -->> Director: "Hello! What can you do?"
+    Director -->> LLM: "Is this a processable question?"
+    LLM -->> Director: "No, return this generic response"
```
harryBedfordSL, line 12: should LLM here be `tasks.py` / something descriptive like `planner` ? Or are we just describing the flow to any LLM

harryBedfordSL: ignore me, I spoke too soon. I can see `taskAgent` below

assets/sum-my-subscriptions/happy_flows.md
```diff
@@ -0,0 +1,83 @@
...
+    Supervisor -->> ComparerTool: ComparerTool.compare('Amazon: £65.15', 'Netflix: £12.99')
+    ComparerTool -->> Supervisor: "Amazon £65.15 is greater than Netflix £12.99 by £52.16"
+```
+
+### Verify we are done
+```mermaid
+sequenceDiagram
+    Supervisor -->> Router: "Find the next best step"
+    Router -->> LLM: "Next best step prompt"
+    LLM -->> Router: "All tasks complete"
```
harryBedfordSL, line 79: are we saying for now that the router does the verification of the answer received ? (Happy just wanted to double check)

harryBedfordSL:
diagrams look awesome! Only thing that I'd love to see is a combination of all of the bottom ones in a generic flow. Feel like we could make use of [loops](https://mermaid.js.org/syntax/sequenceDiagram.html#loops) around the supervisor and executing an agent to show that flow. (This is me thinking out loud, we would obviously lose the context of the problem you're describing being solved, so am unsure what I prefer)

assets/sum-my-subscriptions/happy_flows.md
Ch0live: In my head it does the verification that the current task is complete. And in the scenario there are no more tasks to move on to it should pick an "all tasks complete" option to tell the supervisor to pass the result of all the tasks back to the frontend

philbaxter (commented):
Looks great!.

One comment though, when we call the LLM from the Router and then subsequently from the DatastoreAgent, the sequence diagram gives the impression that its the same LLM. Should it? In practice they may be two different LLM's depending on what we need them to do and cost etc.

harryBedfordSL:
> I was curious if we should
> 
> * add a final summariser method at the top level
> * include the scratchpad here when saving between runs? Might be useful for the summariser I mentioned above

would the summeriser method be an ability of the director? I think we could include the scratchpad if we put the looping element in of going between tasks

Ch0live:
I suppose I didn't distinguish between different kinds of llms, mostly just wanted to capture that a call to some sort of model was being made. I could rename the LLM to LLM.py, which in my head will eventually be designed to do multiple different model calls

Ch0live:
> diagrams look awesome! Only thing that I'd love to see is a combination of all of the bottom ones in a generic flow. Feel like we could make use of [loops](https://mermaid.js.org/syntax/sequenceDiagram.html#loops) around the supervisor and executing an agent to show that flow. (This is me thinking out loud, we would obviously lose the context of the problem you're describing being solved, so am unsure what I prefer)

This would probably let us do another look from an even higher level, I'll add it within another graph file

harryBedfordSL (dismissed):
looks good to me

DannyNicholas (changes requested):
Happy to approve but the black text isn't readable when in "light mode". The dark blue colours clash with the black text. Can you consider modifying the colours to work in both light and dark modes?

Ch0live:
Fixed the black text in light mode by removing the colours (the simplest solution)

assets/sum-my-subscriptions/happy_flows_sum_my_subs_python_modules.md
```diff
@@ -0,0 +1,109 @@
+# "What did I spend more on last month, Amazon or Netflix?"
+
+I am a user who wants to know if last month I spent more on my Netflix subscription than my Amazon one
+
+### Find tasks from question
+```mermaid
+sequenceDiagram
+```
```
harryBedfordSL, line 8: this one meant to be empty?

Ch0live: Nope 😅 removed

harryBedfordSL (approved):

PR #32, opened 2024-04-17
add unhappy path diagrams

harryBedfordSL:
## Description
Adds diagrams for the unhappy paths in the next sum my subscriptions epic

## Changelog
- Adds new Markdown file with mermaid diagrams listed inside

## Notes
- I have added some notes that we can discuss at a later point or now, whichever people prefer. 
- I will keep this PR as a draft until https://github.com/WaitThatShouldntWork/InferGPT/pull/30 is merged, then I will change the target branch

## Diagrams added:

image

image


assets/sum-my-subscriptions/unhappy_flows.md
```diff
@@ -0,0 +1,35 @@
...
+    Router -->> Supervisor: "Call DatastoreAgent with prompt 'get Amazon spending'"
+    Supervisor -->> DatastoreAgent: "Find the answer to the following: 'get Amazon spending'"
+    DatastoreAgent -->> LLM: "Function call: identify methods"
+    LLM -->> DatastoreAgent: "No appropriate method found"
+    DatastoreAgent -->> Supervisor: "I cannot solve this task"
+    Supervisor -->> Router: "Agent failed, try again"
+    Router -->> Supervisor: "No more avaialble agents, task failed"
+    Supervisor -->> Director: "Unable to solve task: 'task_description'"
+    Director -->> frontend: "I'm sorry, I was unable to solve that task"
+```
```
bheyman-scottlogic, line 35: A couple of thoughts: 
1. Is there another path where the agent calls the correct function and then the function fails in some way ie. db connection doesn't work? How would this be handled. Would it try again or just say failed?
2. If one task fails do we move onto the next task or does the whole thing fail? Does the flow need to show all tasks?

bheyman-scottlogic (commented):
Just added a couple of comments, good work though!

assets/sum-my-subscriptions/unhappy_flows.md
bheyman-scottlogic: For the first point, in the sequence diagram there is no flow where a correct method is identified and then still fails ie. carrying on to the database layer (not shown in the diagram) and failing there.

In this scenario do we try the same agent again? Do we try and get better arguments for the query? I feel a level of looping that may happen hasn't been included. Not sure if this detail has been confirmed yet though so happy to discuss

assets/sum-my-subscriptions/unhappy_flows.md
```diff
@@ -0,0 +1,35 @@
...
+sequenceDiagram
+    Router -->> LLM: "Determine which agent is best for this task"
+    LLM -->> Router: "Use DatastoreAgent"
+    Router -->> Supervisor: "Call DatastoreAgent with prompt 'get Amazon spending'"
+    Supervisor -->> DatastoreAgent: "Find the answer to the following: 'get Amazon spending'"
+    DatastoreAgent -->> LLM: "Function call: identify methods"
+    LLM -->> DatastoreAgent: "No appropriate method found"
+    DatastoreAgent -->> Supervisor: "I cannot solve this task"
+    Supervisor -->> Router: "Agent failed, try again"
+    Router -->> Supervisor: "No more avaialble agents, task failed"
```
bheyman-scottlogic, line 32: Typo in "available"

assets/sum-my-subscriptions/unhappy_flows.md
Ch0live: We're comparing the scenario where we fail a task and fail overall and attempt to do a task n times and fail each time, then eventually failing overall. I think the `loop` notation would probably be appropriate here even if it complicates the graph a little bit but also happy to leave as it is as in the end the behaviour is the same.

Ch0live (approved):
Great stuff! Added 1 comment but also happy for this to be merged as it is

assets/sum-my-subscriptions/unhappy_flows.md
harryBedfordSL: good spot, changing now

bheyman-scottlogic (approved):

Ch0live (approved):

PR #37, opened 2024-04-22
Add Neo4J DBMS image to docker compose

Ch0live:
## Description

Adds Neo4J DBMS image to the docker way of running the app. Fixes and updates surrounding tests also

## Changelog
- Add Dockerfile to `/data` that runs `neo4j:5.19.0` image in a new container
- Fixes tests in `api_test.py` and adds ones to test all scenarios (excluding when the backend service is down entirely)
- Updated README to include deletion of containers and images created by `docker compose` command

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,8 @@ Once you have Git installed, follow these steps:
 > [!WARNING]  
 > the first time you do this it may take longer as the compose file builds the images.
 > Backend unittests cannot be run against the docker installation - please follow the [backend README](backend/README.md) for testing guidance
-- View the frontend at [localhost:8650](http://localhost:8650) and the Neo4j Browser as [localhost:7474](http://localhost:8650)
+- View the frontend at [localhost:8650](http://localhost:8650) and the Neo4j Browser at [localhost:7474](http://localhost:7474)
+- use the keyphrase "healthcheck" to test if everything is collected successfully
 
 #### Using pip and python
 - Open your preferred terminal where the project is cloned
--- a/README.md
+++ b/README.md
@@ -46,20 +46,21 @@ Once you have Git installed, follow these steps:
 - Run the git clone command for the fork you just created.
 
 ### Running the service
-- create `.env` files as instructed within the [frontend README](frontend/README.md) and [backend README](backend/README.md)
+- create `.env` files as instructed within the [frontend README](frontend/README.md), [backend README](backend/README.md) and [data README](data/README.md)
 
 #### Using Docker
 - Open your preferred terminal where the project is cloned
 - **Ensure Docker is installed and running**. The easiest way to do this is by using the Docker Desktop app (install it from [docker.com](docker.com) if you don't have it).
 - In the root directory of the project run `docker compose up`
-  - Note the first time you do this it may take longer as the compose file builds the images
-  - to delete and recreate the images use `docker compose down`
+> [!WARNING]  
+> the first time you do this it may take longer as the compose file builds the images.
+> Backend unittests cannot be run against the docker installation - please follow the [backend README](backend/README.md) for testing guidance
+- View the frontend at [localhost:8650](localhost:8650) and the Neo4j Browser as [localhost:7474](localhost:8650)
 
 #### Using pip and python
 - Open your preferred terminal where the project is cloned
-- **Ensure Neo4J Desktop is installed and running**
-  - Instructions can be found [here](https://neo4j.com/docs/desktop-manual/current/installation/)
 - It's recommended (though not technically required) to create a virtual environment for the project by running `python -m venv .venv` to create it and `.venv/Scripts/activate` to activate it in your active terminal.
+- Follow the [data README](data/README.md) for guidance on how to set up Neo4j Desktop
 - Follow the [frontend README](frontend/README.md) to set up and run the front end
 - Follow the [backend README](backend/README.md) to set up and run the back end
  
--- a/README.md
+++ b/README.md
@@ -41,34 +41,30 @@ To clone the repository, you need to have Git installed on your system. Use the 
 
 Once you have Git installed, follow these steps:
 
-- Open your terminal.
+- Open your preferred terminal.
 - Navigate to the directory where you want to clone the repository.
 - Run the git clone command for the fork you just created.
 
-### Install Neo4j Desktop or Aura DB
-Instructions can be found [here](https://neo4j.com/docs/?utm_medium=PaidSearch&utm_source=google&utm_campaign=GDB&utm_content=EMEA-X-Conversion-GDB-Text&utm_term=neo4j&gclid=Cj0KCQiA1rSsBhDHARIsANB4EJY8wQONKSyNCofQBGAcOGWwNpNh4Z0yj7oGxok8vs2CipPJMjGPcpkaAuw1EALw_wcB).  
-Install the pre-made recommendations database. Version 4.4 is the latest stable.
-
 ### Running the service
-- Create `.env` files based on the `.env.example` files under `/frontend` and `/backend`
+- create `.env` files as instructed within the [frontend README](frontend/README.md) and [backend README](backend/README.md)
 
 #### Using Docker
-- Start and open the Neo4J Desktop app. Start the default database (neo4j).
-- create `.env` files as instructed within the [frontend README](frontend/README.md) and [backend README](backend/README.md)
-- Run the Docker Desktop app (install it from [docker.com](docker.com) if you don't have it)
+- Open your preferred terminal where the project is cloned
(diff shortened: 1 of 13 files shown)


Ch0live:
To test this run

- `docker compose up` to run all 3 containers
- `pytest` within the `/backend` directory to check all tests pass

image

compose.yml
```diff
@@ -1,7 +1,10 @@
 version: '1'
```
philbaxter, line 1: My docker version (26.0.0) complains about `version` being obsolete. Not an error, just a warning

compose.yml
```diff
@@ -1,7 +1,10 @@
...
 services:
-  # neo4j-store:
-  #   build: "./data/"
+  neo4j-dbms:
+    build: "data/"
+    ports:
+      - 7474:7474
+      - 7687:7687
```
philbaxter, line 7: I already have a dockerize neo4j running which conflicts with these ports, perhaps we should use non-standard ports within this compose file

compose.yml
```diff
@@ -1,7 +1,10 @@
 version: '1'
 services:
-  # neo4j-store:
-  #   build: "./data/"
+  neo4j-dbms:
```
philbaxter, line 3: We should set the default username/password for this instance, maybe drag the values from an env file?

We should also create a volume for neo4j data so that we can persist our graphs between invocations of docker compose up/down

philbaxter (changes requested):

hsauve-scottlogic (approved):
Looks good Charlie, nice bit of tidying up too. I just had one question about deleting docker images

DannyNicholas:
Pulled in the latest version but the tests fail...

```
INFO     api:api.py:36 health_check method called
INFO     root:graph_db_utils.py:14 testing database connection...
CRITICAL root:graph_db_utils.py:22 connection connection failed
CRITICAL root:graph_db_utils.py:23 Couldn't connect to host.docker.internal:7687 (resolved to ()):
Timed out trying to establish connection to ResolvedIPv4Address(('10.12.4.111', 7687))
INFO     root:api.py:40 health_check method failed - neo4j connection unsuccessful
INFO     root:api.py:46 health_check method complete
INFO     httpx:_client.py:1013 HTTP Request: GET http://testserver/health "HTTP/1.1 500 Internal Server Error"
=========================================================================================================== short test summary info ============================================================================================================ 
FAILED api_test.py::test_health_check_response_healthy - assert 500 == 200
```

I can get the tests to pass by changing...
```
NEO4J_URI="bolt://host.docker.internal:7687"
```
to
```
NEO4J_URI="bolt://localhost:7687"
```
...but then the app won't run as the backend can no longer connect to neo4j.

README.md
```diff
@@ -41,33 +41,33 @@ To clone the repository, you need to have Git installed on your system. Use the
...
 
 #### Using Docker
-- Start and open the Neo4J Desktop app. Start the default database (neo4j).
-- create `.env` files as instructed within the [frontend README](frontend/README.md) and [backend README](backend/README.md)
-- Run the Docker Desktop app (install it from [docker.com](docker.com) if you don't have it)
+- Open your preferred terminal where the project is cloned
+- **Ensure Docker is installed and running**. The easiest way to do this is by using the Docker Desktop app (install it from [docker.com](docker.com) if you don't have it).
 - In the root directory of the project run `docker compose up`
   - Note the first time you do this it may take longer as the compose file builds the images
+  - to delete and recreate the images use
```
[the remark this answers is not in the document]
Ch0live: No, I just had some experiences when changing the configuration that the changes weren't transferred unless I recreated the image. If you're only making code changes they should update without needing to recreate the image

compose.yml
Ch0live: Removed

hsauve-scottlogic (approved):

compose.yml
Ch0live: username/password defined in the .env file under /data now 👍

compose.yml
Ch0live: This will require changing the image which from some initial research looks like a lot of work. I'd say it's better for us (at least for now) to just stop the databases we're running locally so we can use the pure image for neo4j

hsauve-scottlogic (dismissed):
Looks very good to me Charlie, I just added some questions, rather than comments :)

backend/api_test.py
```diff
@@ -23,13 +31,12 @@ def test_chat_success_response(mocker):
     assert response.json() == expected_message
 
 
-def test_chat_failure_response(mocker):
-    expected_message = "InferGPT backend is healthy"
+def test_chat_response(mocker):
```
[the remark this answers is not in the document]
Ch0live: Yes! Good spot, changed

data/README.md
```diff
@@ -4,9 +4,15 @@ A directory for all util and service modules for storing, manipulating and retri
 
 InferGPT uses a Neo4j Knowledge Graph exclusively.
 
+# Setup
+
+1. Create .env files. There are template files (.env.example) for you to copy with comments for guidance.
+2. If you're running the service locally, set up Neo4j Desktop as guided in the section below.
+If you're running the service using Docker, ignore this step.
```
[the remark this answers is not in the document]
Ch0live: Docker is easier to get the service up and running as it removes the need to set up neo4j desktop entirely. Running stuff locally I think will reload frontend/backend faster than Docker however.

Both are valid and equally useful, but the Docker option is especially helpful to newcomers as it removes a lot of extra steps included in running all the individual parts

hsauve-scottlogic (approved):

DannyNicholas:
I am still getting test failures when running `pytest`.

```
CRITICAL root:graph_db_utils.py:22 database connection failed
CRITICAL root:graph_db_utils.py:23 Couldn't connect to host.docker.internal:7687 (resolved to ()):
Timed out trying to establish connection to ResolvedIPv4Address(('10.12.4.111', 7687))
INFO     root:api.py:40 health_check method failed - neo4j connection unsuccessful
INFO     root:api.py:46 health_check method complete
INFO     httpx:_client.py:1013 HTTP Request: GET http://testserver/health "HTTP/1.1 500 Internal Server Error"
============================================================== short test summary info ==============================================================
FAILED api_test.py::test_health_check_response_healthy - assert 500 == 200
```

Should `api_test.py` use mocks now when calling neo4j?

philbaxter (dismissed):

Ch0live:
@DannyNicholas yes they should, I think pytest should only be run with the local run set up and not against the service ran in Docker.

Ideally as we spoke before it should be agnostic to the service running entirely so is no longer a concern here

Ch0live:
If you agree @DannyNicholas let me know by approving or commenting and I'll merge if you approve additionally

DannyNicholas:
My view is that the tests should pass regardless of what's in the `.env` file. We don't want to force people to have to change their `.env` file every time they want to run the tests.

In this case, the test results in calls to `test_connection()` causing `driver.verify_connectivity()` to make real network calls. My view is that we should avoid real network calls in a unit test.

In other tests were mocking `api.test_connection`,  so I think we need to add the same mocking behaviour to the healthy response test...

```
def test_health_check_response_healthy(mocker):

    mock_connection_test = mocker.patch("api.test_connection", return_value=True)

    response = client.get("/health")

    mock_connection_test.assert_called()

    assert response.status_code == 200
    assert response.json() == healthy_response
```

This avoids us testing the real connection to neo4j so ensures the test passes. Happy to discuss if I've misunderstood the intentions.

backend/api_test.py
```diff
@@ -1,19 +1,27 @@
...
 utterance = "Hello there"
+expected_message = "Hello to you too! From InferGPT"
 
-
-def test_health_check_response():
+def test_health_check_response_healthy():
     response = client.get("/health")
 
     assert response.status_code == 200
     assert response.json() == healthy_response
```
DannyNicholas, line 12: The failing test is `test_health_check_response`. The test results in real calls to `test_connection()` causing `driver.verify_connectivity()` to make real network calls. As a result, the test only passes if neo4j is running (and using the URL/credentials provided in the `.env` file).

We should avoid real network calls in a unit tests. In other tests within `api.py` we're mocking `api.test_connection`,  so I think we need to add the same mocking behaviour to this healthy response test. This makes our test consistent with the other tests and ensures we're only testing the logic in `api.py` (as well as ensuring the test passes):

```
def test_health_check_response_healthy(mocker):

    mock_connection_test = mocker.patch("api.test_connection", return_value=True)

    response = client.get("/health")

    mock_connection_test.assert_called()

    assert response.status_code == 200
    assert response.json() == healthy_response
```

Ch0live: Yes, no you're right. Happy to fix the tests to uncouple them from the service running here 👍

Ch0live: I have mocked out the `test_connection()` call in the latest commit

image

DannyNicholas (approved):

philbaxter (approved):

PR #38, opened 2024-04-23
Create tasks from question

hsauve-scottlogic:
## Description

A goal/intention is now broken down into up to 5 tasks

## Changelog
- Created a create_tasks_agent that breakdowns a question into up to 5 tasks
- Updated director code to call create_tasks_agent

@bheyman-scottlogic


backend/agents/create_tasks_agent.py
```diff
@@ -0,0 +1,41 @@
...
+
+    EG. prompt: "How much did I spend last month on my debit card?"
+    1) Find all spending transactions last month on the user's debit card. Explanation: {your reasoning}
+    2) Sum all transactions. Explanation: {your reasoning}
+"""
+
+def create_tasks(user_prompt):
+    logger.info("create_tasks function is called")
+    response = call_model(create_tasks_prompt, user_prompt)
+    return(response)
```
harryBedfordSL, line 41: is there a reason why we're using a tuple as the return type here? Reading the method name `create_tasks` makes me think some sort of array / list will be returned

backend/director.py
```diff
@@ -25,17 +42,15 @@
 
 
 def question(question):
```
philbaxter, line 44: should we specify return type here aswell? str?

backend/director.py
```diff
@@ -25,17 +42,15 @@
...
+        return create_tasks(question)
 
-    return call_model(system_prompt, user_prompt=question)
+    logger.info("director calling call_model function")
+    return call_model(director_prompt, user_prompt=question)
 
 
-def determine_intent(question):
-    return call_model(system_prompt_to_determine_intent, user_prompt=question)
+def determine_intention(question):
```
philbaxter, line 54: similarly specify return type?

philbaxter (changes requested):
I know return types can be inferred, but if we specify them in one place, we should most likely be consistent and do it in the others too

philbaxter (approved):

harryBedfordSL (approved):

PR #39, opened 2024-04-26
Design agent class

harryBedfordSL:
## Description

This PR creates the `Agent` base class and implements the `data_store.py` agent. Note, currently the implemented agent is not wired up to the neo4j database or the supervisor. A test endpoint was added during development to test the logic with a default answer, `£6.99` always being returned from the agent's tool. Multiple tools were tested within the agent too, but just one has been left as it aligns with the epic

## Changelog
- Created Agent abstract class
- Created implementation of agent, `data_store.py`
- Created tool adapter for mistral

## Manual testing

To wire up and use the `data_store.py` agent I was making the following changes to the two files:

Bottom of `data_store.py` :
```python
def run():
    db_agent = DatastoreAgent()
    answer = db_agent.invoke("How much did I spend on Spotify last month?")
    print(f"ANSWER: {answer}")
```

Add new endpoint to `api.py`
```python
from agents.data_store import run

...

@app.get("/test")
async def test():
    run()
    return JSONResponse(status_code=200, content=healthy_response)
```

then just use postman (or a browser!) to hit http://127.0.0.1:8250/test

### DIFF
--- a/backend/agents/adapters.py
+++ b/backend/agents/adapters.py
@@ -1,0 +1,26 @@
+from .tool import Tool
+
+
+def get_mistral_tool(tool: Tool):
+    return {
+        "type": "function",
+        "function": {
+            "description": tool.description,
+            "name": tool.action.__name__,
+            "parameters": {
+                "type": "object",
+                "properties": {
+                    key: { 
+                        "type": inner_dict.type,
+                        "description": inner_dict.description,
+                    }
+                    for key, inner_dict in tool.parameters.items()
+                },
+                "required": [
+                    key
+                    for key, inner_dict in tool.parameters.items()
+                    if inner_dict.required
+                ],
+            },
+        },
+    }
--- a/backend/agents/adapters.py
+++ b/backend/agents/adapters.py
@@ -10,17 +10,13 @@ def get_mistral_tool(tool: Tool):
             "parameters": {
                 "type": "object",
                 "properties": {
-                    key: { 
+                    key: {
                         "type": inner_dict.type,
                         "description": inner_dict.description,
                     }
                     for key, inner_dict in tool.parameters.items()
                 },
-                "required": [
-                    key
-                    for key, inner_dict in tool.parameters.items()
-                    if inner_dict.required
-                ],
+                "required": [key for key, inner_dict in tool.parameters.items() if inner_dict.required],
             },
         },
     }
--- a/backend/agents/adapters.py
+++ b/backend/agents/adapters.py
@@ -1,4 +1,4 @@
-from .tool import Tool
+from agents import Tool
 
 
 def get_mistral_tool(tool: Tool):
--- a/backend/agents/adapters.py
+++ b/backend/agents/adapters.py
@@ -1,7 +1,7 @@
-from agents import Tool
+from .tool import Tool
 
 
-def get_mistral_tool(tool: Tool):
+def convert_to_mistral_tool(tool: Tool):
     return {
         "type": "function",
         "function": {
--- a/backend/agents/adapters_test.py
+++ b/backend/agents/adapters_test.py
@@ -1,4 +1,6 @@
-from agents import tool_metadata, get_mistral_tool, Parameter
+from agents import tool_metadata
+from .adapters import convert_to_mistral_tool
+from .types import Parameter
 
 name = "Mock Tool"
 description = "A test tool"
@@ -34,4 +36,4 @@ def test_tool_conversion():
         },
     }
 
-    assert get_mistral_tool(mock_tool) == expected_output
+    assert convert_to_mistral_tool(mock_tool) == expected_output
--- a/backend/agents/adapters_test.py
+++ b/backend/agents/adapters_test.py
@@ -1,0 +1,47 @@
+from agents.tool import tool_metadata
+from agents.adapters import get_mistral_tool
+from agents.types import Parameter
+
+name = "Mock Tool"
+description = "A test tool"
+param_description = "A string"
+
+@tool_metadata(
+    name=name,
+    description=description,
+    parameters={
+        "input": Parameter(
+            type="string",
+            description=param_description,
+            required=True
+        ),
+        "optional": Parameter(
+            type="string",
+            description=param_description,
+            required=False
+        )
+    },
+
+)
+def mock_tool(input: str):
+    return input 
+
+
+def test_tool_conversion():
+    expected_output = {
+        "type": "function",
+        "function": {
+            "description": description,
+            "name": "mock_tool",
+            "parameters": {
+                "type": "object",
+                "properties": {
+                    "input": { "type": "string", "description": param_description }, 
+                    "optional": { "type": "string", "description": param_description }, 
+                },
+                "required": ["input"],
+            },
+        },
+    }
+
+    assert get_mistral_tool(mock_tool) == expected_output
--- a/backend/agents/adapters_test.py
+++ b/backend/agents/adapters_test.py
@@ -1,6 +1,4 @@
-from agents.tool import tool_metadata
-from agents.adapters import get_mistral_tool
-from agents.types import Parameter
+from agents import tool_metadata, get_mistral_tool, Parameter
 
 name = "Mock Tool"
 description = "A test tool"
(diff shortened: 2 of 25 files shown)


backend/agents/data_store.py
```diff
@@ -0,0 +1,38 @@
...
+    parameters={
+        "merchant": Parameter(
+            type="string",
+            description="Merchant name eg. Amazon",
+        )
+    },
+)
+def get_db_data(merchant: str) -> str:
+    print(f"Getting data from database with merchant: {merchant}")
+    return "£6.99"
```
philbaxter, line 18: I know this is a stub, but considering we may be `summing` these values, should we just return a raw number and suppress the currency sign?. I guess this also has a bearing on how we store the data in the graph.

backend/agents/data_store.py
```diff
@@ -0,0 +1,38 @@
+from .tool import toolMetadata
+from .types import Parameter
+from .agent import Agent, agentMetadata
+
+
+@toolMetadata(
+    name="Get Database Data",
+    description="This tool gets data from the database",
```
DannyNicholas, line 8: Since this tool has a specific function (until we build the ability for generic queries), a more specific name and description may help the LLM to choose this function appropriately. "gets data" may suggest it can get any data from the database that may cause problems with certain requests.

Since transactions is now part of our ontology, should our description be something like "This tool gets the sum of all transactions for a specified merchant"?

I guess it depends on what types of database requests we expect this function to handle. Please ignore if this is intentionally generic for now.

backend/agents/data_store.py
harryBedfordSL: ooooo good question! Instinctively, I agree with you. But at the moment this method is a tool, which is defined as a callable method that returns a string. I feel like everything will need to end up back as a string at some point, the question being what thing should do the conversion back to a string? 

Or should I just change this to `"6.99"` for now?

DannyNicholas: Great questions..... It's not clear in my head what is the best return type to allow the next step to take this value and process it successfully. Feels like we need some kind of model/structure.

I'm happy to just use `"6.99"` for now, although long-term we'd need to know the currency type.

One option is that we return a dictionary that holds `amount` and `currency_code` or encode those values into a JSON string?

Shall we discuss on Monday how we future-proof return types for agent functions?

backend/agents/data_store.py
harryBedfordSL: Yeah was keeping it deliberately generic for now. More than happy for this to be changed as the agent logic is implemented more

DannyNicholas (approved):
Nice work Harry - great looking code.

Had to use Copilot for some bits of python I didn't quite understand. 😄

Ch0live:
> Manual testing
> To wire up and use the data_store.py agent I was making the following changes to the two files:
> 
> Bottom of data_store.py :
> 
> def run():
    db_agent = DatastoreAgent()
    answer = db_agent.invoke("How much did I spend on Spotify last month?")
    print(f"ANSWER: {answer}")
> Add new endpoint to api.py
> 
> from agents.data_store import run
> 
> ...
> 
> @app.get("/test")
> async def test():
    run()
    return JSONResponse(status_code=200, content=healthy_response)
> then just use postman (or a browser!) to hit http://127.0.0.1:8250/test


Guessing this has been removed?

backend/agents/adapters_test.py
```diff
@@ -0,0 +1,39 @@
+from agents.tool import tool_metadata
```
Ch0live, line 1: If you add these method names to the `__init__.py` we can import all of this in 1 line
```python
from agents import tool_metadata, get_mistral_tool, Parameter
```

backend/agents/agent.py
```diff
@@ -0,0 +1,38 @@
+from abc import ABC
+from typing import List, Type
+from utils import call_model_with_tools
+from .adapters import get_mistral_tool
```
Ch0live, line 4: [Import comment from before](https://github.com/WaitThatShouldntWork/InferGPT/pull/39/files#r1584566963)

backend/agents/agent.py
```diff
@@ -0,0 +1,38 @@
...
+    prompt: str
+
+    def __get_action(self, utterance: str) -> Action_and_args:
+        tools = map(get_mistral_tool, self.tools)
+        (function_name, function_params) = call_model_with_tools(self.prompt, utterance, tools)
+
+        tool = next((tool for tool in self.tools if tool.action.__name__ == function_name), None)
+
+        if tool is None:
+            raise ValueError(f"Tool {function_name} not found in agent {self.name}")
```
Ch0live, line 21: One for the future probably, but if we fail to pick a tool we should look at doing a retry LLM call with a log of what went wrong to see if it can resolve it's issue

backend/agents/types.py
```diff
@@ -0,0 +1,24 @@
+from typing import Callable, Protocol, Tuple, TypeVar
```
Ch0live, line 1: I wasn't aware Python had this `lib/typing` module. I've probably used it under the hood a bunch of times but not directly pulled it into a module like this. Neat!

backend/api.py
```diff
@@ -12,6 +12,9 @@
 app = FastAPI()
 config = Config()
 
+if config.frontend_url is None:
```
Ch0live, line 15: Why are we checking for just the `frontend_url` property? Also this is the only property I gave a default value

Ch0live (changes requested):
Left some comments, but great job on this. Nice Python classes and decorator 🐍

harryBedfordSL:
> > Manual testing
> > To wire up and use the data_store.py agent I was making the following changes to the two files:
> > Bottom of data_store.py :
> > def run():
> > db_agent = DatastoreAgent()
> > answer = db_agent.invoke("How much did I spend on Spotify last month?")
> > print(f"ANSWER: {answer}")
> > Add new endpoint to api.py
> > from agents.data_store import run
> > ...
> > @app.get("/test")
> > async def test():
> > run()
> > return JSONResponse(status_code=200, content=healthy_response)
> > then just use postman (or a browser!) to hit http://127.0.0.1:8250/test
> 
> Guessing this has been removed?

yeah exactly, not needed in the end feature, but just how I tested it whilst developing

backend/agents/adapters_test.py
harryBedfordSL: like it, changing

backend/agents/agent.py
harryBedfordSL: yeah I like it, will change. Although as I am doing this, is there a subtle difference between what should be in the init file and locally imported i.e. is the init file exposing to other modules what is available from this one? Whereas there maybe some functionality that we don't want other modules to use? (me thinking out loud, happy to change to init file approach for now with all as like it)

backend/agents/agent.py
harryBedfordSL: agree, and agree yeah one for the future.  Should be nice to change in here though and will apply for all agents!

backend/agents/adapters.py
```diff
@@ -0,0 +1,22 @@
+from .tool import Tool
+
+
+def get_mistral_tool(tool: Tool):
```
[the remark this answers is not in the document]
Ch0live: Ahh I see it now! We store tools for the Agent as a Tool and then convert _those tools_ into the JSON format of the mistral function call tools.

I didn't see that at first, maybe it should be `convert_to_mistral_function_calling_tools`, `map_to_mistral_function_calling_tools` or just `map_to_mistral_tool`?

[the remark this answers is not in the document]
Ch0live: If we wanna leave it as it is happy to approve with this name too. Just throwing ideas into the mix

backend/agents/agent.py
Ch0live: This is true and something I wasn't thinking about, if outside the /agents directory it shouldn't be accessible then lets keep it out of the __init__ file and import relatively (as that's appropriate as all the files that want that method are _relatively_ close to it)

Ch0live (dismissed):

Ch0live (approved):

DannyNicholas (approved):

PR #40, opened 2024-04-26
Add jinja2 templates

Ch0live:
## Description

Adds templating for creating complex prompts with insertable variables. Plus a `prompting.py` module containing a class `PromptEngine` to make loading the templates easy. Code changes based on AutoGPT architecture

## Changelog
- Adds `prompting.py` for easy loading of the templates anywhere in the codebase
- Tested using `prompting.py`
- Adds initial templates required for the router
- Removes `version` from docker `compose.yml` (deprecated field)


backend/prompts/prompting.py
```diff
@@ -0,0 +1,48 @@
+import logging
+import os
+
+from jinja2 import Environment, FileSystemLoader
+
+
+class PromptEngine:
+    """
+    Class to handle loading and populating Jinja2 templates for prompts.
+    """
```
bheyman-scottlogic, line 10: We haven't used docstrings anywhere else. Should be consistent across the board. Either by adding them in or removing them here

backend/prompts/prompting.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def __init__(self, model: str):
+        """
+        Initialize the PromptEngine with the specified model.
+
+        Args:
+            model (str): The model to use for loading prompts.
+            debug_enabled (bool): Enable or disable debug logging.
```
harryBedfordSL, line 18: looks like `debug_enabled` isn't used anywhere?

backend/prompts/prompting_test.py
```diff
@@ -0,0 +1,21 @@
...
+    engine = PromptEngine("mistral")
+    try:
+        engine.load_prompt("task-step")
+    except Exception:
+        raise
+
+def test_load_agent_selection_format_template():
+    engine = PromptEngine("mistral")
+    try:
+        engine.load_prompt("agent-selection-format")
```
harryBedfordSL, line 19: is there an assertion we can do here and for the other tests? (Forgive me if I'm missing it)

backend/prompts/prompting.py
Ch0live: True, I've removed it from the init method signature. Removed

backend/prompts/prompting.py
Ch0live: I agree, removed also

backend/prompts/prompting_test.py
Ch0live: I've added full string assertions

harryBedfordSL (approved):
looks great to me

bheyman-scottlogic (approved):
Good stuff Charlie!

PR #41, opened 2024-04-29
Create unit tests for Graph DB Utils

DannyNicholas:
## Description

This PR creates a new test for `graph_db_utils.py` that covers the `verify_connectivity()` and `create_goal()` methods.

The tests mock the neo4j `driver` and `session` to verify the expected behaviour when `graph_db_utils.py` uses these neo4j instances.

Test cover:

- Confirm  `verify_connectivity()` returns `True` when we can connect to neo4j database
- Confirm  `verify_connectivity()` returns `False` when the connection to neo4j database fails
- Confirm  `create_goal()` results in calls to the neo4j `session.run()` method to create a goal node
- Confirm  `create_goal()` throws an exception when the neo4j `session.run()` fails

## Changelog

-  Created test file `graph_db_utils_test.py`


backend/utils/graph_db_utils_test.py
```diff
@@ -0,0 +1,52 @@
+from unittest.mock import ANY, MagicMock
+from neo4j import Driver, Session
+import pytest
+import utils.graph_db_utils
+
+def test_database_connectivity_is_healthy(mocker):
+    mock_driver = mocker.patch("utils.graph_db_utils.driver", return_value=MagicMock(spec=Driver))
+    mock_driver.verify_connectivity.return_value = None
+    
+    connectivity = utils.graph_db_utils.test_connection()
```
DannyNicholas, line 10: I've had to fully-qualify the calls to the methods being tested.

Initially, I tried the import below...
```
from utils.graph_db_utils import test_connection
```
... but this resulted in the `PyTest` warning:
```
Expected None, but utils/graph_db_utils_test.py::test_connection returned False,
which will be an error in a future version of pytest
```
... also this resulted in the mock driver being ignored and the test trying to make real HTTP calls to the neo4j database.

DannyNicholas:
Replaced by https://github.com/WaitThatShouldntWork/InferGPT/pull/42

PR #42, opened 2024-04-29
Create unit tests for Graph DB Utils

DannyNicholas:
## Description

This PR creates a new test for `graph_db_utils.py` that covers the `verify_connectivity()` and `create_goal()` methods.

The tests mock the neo4j `driver` and `session` to verify the expected behaviour when `graph_db_utils.py` uses these neo4j instances.

Test cover:

- Confirm  `verify_connectivity()` returns `True` when we can connect to neo4j database
- Confirm  `verify_connectivity()` returns `False` when the connection to neo4j database fails
- Confirm  `create_goal()` results in calls to the neo4j `session.run()` method to create a goal node
- Confirm  `create_goal()` throws an exception when the neo4j `session.run()` fails

## Changelog

-  Created test file `graph_db_utils_test.py`.

### DIFF
--- a/backend/utils/graph_db_utils_test.py
+++ b/backend/utils/graph_db_utils_test.py
@@ -1,0 +1,52 @@
+from unittest.mock import ANY, MagicMock
+from neo4j import Driver, Session
+import pytest
+import utils.graph_db_utils
+
+def test_database_connectivity_is_healthy(mocker):
+    mock_driver = mocker.patch("utils.graph_db_utils.driver", return_value=MagicMock(spec=Driver))
+    mock_driver.verify_connectivity.return_value = None
+
+    connected = utils.graph_db_utils.test_connection()
+
+    assert connected
+    mock_driver.verify_connectivity.assert_called_once()
+    mock_driver.close.assert_called_once()
+
+
+def test_database_connectivity_is_unhealthy(mocker):
+    mock_driver = mocker.patch("utils.graph_db_utils.driver", return_value=MagicMock(spec=Driver))
+    mock_driver.verify_connectivity.side_effect = Exception
+
+    connected = utils.graph_db_utils.test_connection()
+
+    assert not connected
+    mock_driver.verify_connectivity.assert_called_once()
+    mock_driver.close.assert_called_once()
+
+
+def test_create_goal_is_successful(mocker):
+    mock_driver = mocker.patch("utils.graph_db_utils.driver", return_value=MagicMock(spec=Driver))
+    mock_session = MagicMock(spec=Session)
+    mock_driver.session.return_value = mock_session
+
+    response = utils.graph_db_utils.create_goal("Test Name", "Test Description")
+
+    assert response is None
+    mock_session.run.assert_called_once_with(ANY, name="Test Name", description="Test Description")
+    mock_session.close.assert_called_once()
+    mock_driver.close.assert_called_once()
+
+
+def test_create_goal_throws_exception(mocker):
+    mock_driver = mocker.patch("utils.graph_db_utils.driver", return_value=MagicMock(spec=Driver))
+    mock_session = MagicMock(spec=Session)
+    mock_driver.session.return_value = mock_session
+    mock_session.run.side_effect = Exception
+
+    with pytest.raises(Exception):
+        utils.graph_db_utils.create_goal("Test Name", "Test Description")
+
+    mock_session.run.assert_called_once_with(ANY, name="Test Name", description="Test Description")
+    mock_session.close.assert_called_once()
+    mock_driver.close.assert_called_once()
--- a/backend/utils/graph_db_utils_test.py
+++ b/backend/utils/graph_db_utils_test.py
@@ -18,8 +18,8 @@ def mock_driver(mocker, mock_session):
 
 def remove_whitespace_and_newlines(original):
     return " ".join(original.replace(r'\n', ' ').replace(r'\r', '').split())
-    
-    
+
+
 def test_database_connectivity_is_healthy(mock_driver):
     mock_driver.verify_connectivity.return_value = None
 
--- a/backend/utils/graph_db_utils_test.py
+++ b/backend/utils/graph_db_utils_test.py
@@ -1,4 +1,4 @@
-from unittest.mock import ANY, MagicMock
+from unittest.mock import MagicMock
 from neo4j import Driver, Session
 import pytest
 # We assign an alias to "test_connection" to avoid pytest treating it as another test function
@@ -48,7 +48,7 @@ def test_create_goal_is_successful(mock_driver, mock_session):
     assert kwargs == { "name":"Test Name", "description":"Test Description" }
     actual_cypher_query = remove_whitespace_and_newlines(str(args[0]))
     expected_cypher_query = "MERGE (g:Goal {name: $name, description: $description}) RETURN g"
-    assert expected_cypher_query == actual_cypher_query
+    assert actual_cypher_query == expected_cypher_query
     mock_session.close.assert_called_once()
     mock_driver.close.assert_called_once()
 
@@ -59,6 +59,6 @@ def test_create_goal_throws_exception(mock_driver, mock_session):
     with pytest.raises(Exception):
         create_goal("Test Name", "Test Description")
 
-    mock_session.run.assert_called_once_with(ANY, name="Test Name", description="Test Description")
+    mock_session.run.assert_called_once()
     mock_session.close.assert_called_once()
     mock_driver.close.assert_called_once()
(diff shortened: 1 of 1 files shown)


backend/utils/graph_db_utils_test.py
```diff
@@ -0,0 +1,52 @@
+from unittest.mock import ANY, MagicMock
+from neo4j import Driver, Session
+import pytest
+import utils.graph_db_utils
+
+def test_database_connectivity_is_healthy(mocker):
+    mock_driver = mocker.patch("utils.graph_db_utils.driver", return_value=MagicMock(spec=Driver))
```
DannyNicholas, line 7: I've had to fully-qualify the calls to the methods being tested.

Initially, I tried the import below...
```
from utils.graph_db_utils import test_connection
```
... but this resulted in the `PyTest` warning:
```
Expected None, but utils/graph_db_utils_test.py::test_connection returned False,
which will be an error in a future version of pytest
```
... also this resulted in the mock driver being ignored and the test trying to make real HTTP calls to the neo4j database.

backend/utils/graph_db_utils_test.py
```diff
@@ -0,0 +1,52 @@
...
+    mock_driver.verify_connectivity.return_value = None
+
+    connected = utils.graph_db_utils.test_connection()
```
DannyNicholas, line 10: I've had to fully-qualify the calls to the methods being tested.

Initially, I tried the import below...
```
from utils.graph_db_utils import test_connection
```
... but this resulted in the `PyTest` warning:
```
Expected None, but utils/graph_db_utils_test.py::test_connection returned False,
which will be an error in a future version of pytest
```
... also this resulted in the mock driver being ignored and the test trying to make real HTTP calls to the neo4j database.

harryBedfordSL: hmm that is strange. Does the warning go away if you change the assertion on line 12 to:

```python
assert connected is False
```

DannyNicholas: @harryBedfordSL  - it appears to be purely triggered by the `import`.

If I comment out all the code and just leave a single import...
```
from utils.graph_db_utils import test_connection
```
...then I get the warning.

backend/utils/graph_db_utils_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+    assert connected
+    mock_driver.verify_connectivity.assert_called_once()
+    mock_driver.close.assert_called_once()
+
+
+def test_database_connectivity_is_unhealthy(mocker):
+    mock_driver = mocker.patch("utils.graph_db_utils.driver", return_value=MagicMock(spec=Driver))
```
harryBedfordSL, line 18: this mock_driver line could be extracted to its own fixture if you wanted to, something like:

```python
@pytest.fixture
def mock_driver(mocker):
  return mocker.patch("utils.graph_db_utils.driver", return_value=MagicMock(spec=Driver))
```

and then `mock_driver` would be available as an arg to each test method, I think... 😅

backend/utils/graph_db_utils_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+def test_create_goal_is_successful(mocker):
+    mock_driver = mocker.patch("utils.graph_db_utils.driver", return_value=MagicMock(spec=Driver))
+    mock_session = MagicMock(spec=Session)
+    mock_driver.session.return_value = mock_session
+
+    response = utils.graph_db_utils.create_goal("Test Name", "Test Description")
+
+    assert response is None
+    mock_session.run.assert_called_once_with(ANY, name="Test Name", description="Test Description")
```
harryBedfordSL, line 36: Should we use something more narrow than `ANY` ? Can we assert against a value?

backend/utils/graph_db_utils_test.py
DannyNicholas: Fixed. The problem was caused by `test_connection`. `pytest` thought this was another test (due to the name).

We now use an alias of `verify_connection` to work-around this,

backend/utils/graph_db_utils_test.py
DannyNicholas: Great tip @harryBedfordSL - done. 👍

backend/utils/graph_db_utils_test.py
DannyNicholas: It's tricky to do this within `assert_called_once_with` as we'd need to exactly match the generated cypher query string, which has a combination of new lines and whitespace. I found it hard to exactly replicate this string in a test, which is why I used `ANY`.

Instead, I've used `call_args` to extract the arguments passed to the method. I can then remove the random whitespace and newlines and do the assertions in separate statements.

harryBedfordSL (dismissed):
Looks great to me

backend/utils/graph_db_utils_test.py
```diff
@@ -0,0 +1,64 @@
+from unittest.mock import MagicMock
+from neo4j import Driver, Session
+import pytest
+# We assign an alias to "test_connection" to avoid pytest treating it as another test function
+from utils.graph_db_utils import test_connection as verify_connection, create_goal
```
Ch0live, line 5: I don't believe you need the `.graph_db_utils` as we have mentioned these methods in the __init__ file
```suggestion
from utils import test_connection as verify_connection, create_goal
```

Ch0live, line 5: Is there a reason we're changing the `test_connection` name to `verify_connection`? If we think it should be renamed I'd prefer we rename it in `graph_db_utils.py` rather than adjust it's name just here

DannyNicholas: Thanks @Ch0live - that works. 👍  Updated!

Ch0live: Interesting, that is a funny issue with `pytest`. With that context I'm happy with either the alias or calling it `verify_connection`

backend/utils/graph_db_utils_test.py
```diff
@@ -0,0 +1,64 @@
...
+@pytest.fixture
+def mock_session():
+    return MagicMock(spec=Session)
+
+
+@pytest.fixture
+def mock_driver(mocker, mock_session):
+    mock_driver = mocker.patch("utils.graph_db_utils.driver", return_value=MagicMock(spec=Driver))
+    mock_driver.session.return_value = mock_session
+    return mock_driver
```
Ch0live, line 16: Clever mocking this. As I understand it, the mock_session and mock_driver methods return Session and Driver MagicMock objects. And we also chain the session mock to the driver mock that then stubs any imports from utils?

backend/utils/graph_db_utils_test.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def remove_whitespace_and_newlines(original):
+    return " ".join(original.replace(r'\n', ' ').replace(r'\r', '').split())
+
+
+def test_database_connectivity_is_healthy(mock_driver):
```
Ch0live, line 23: Not sure if it's more readable but technically we could shorten these tests like so
```suggestion
def test_database_connectivity_is_healthy(mock_driver):
    mock_driver.verify_connectivity.return_value = None

    assert verify_connection()
    mock_driver.verify_connectivity.assert_called_once()
```

Again, your pick, I think the current version is more readable, just creates a variable that is used once

backend/utils/graph_db_utils_test.py
```diff
@@ -0,0 +1,64 @@
...
+    mock_driver.verify_connectivity.return_value = None
+
+    connected = verify_connection()
```
Ch0live, line 26: Not sure if it's more readable but technically we could shorten these tests like so
```suggestion
def test_database_connectivity_is_healthy(mock_driver):
    mock_driver.verify_connectivity.return_value = None

    assert verify_connection()
    mock_driver.verify_connectivity.assert_called_once()
```

Again, your pick, I think the current version is more readable, just creates a variable that is used once

Ch0live (approved):
Good stuff! Approved as my suggested changes are just suggestions 👍

backend/utils/graph_db_utils_test.py
DannyNicholas: Exactly. 👍  If you pull-in the `mock_driver` into a test, then any calls to `driver.session()` will return the `mock_session`.

backend/utils/graph_db_utils_test.py
DannyNicholas: It's a good point.

In my case, I'm attempting to follow the "arrange, act, assert" approach with a clear separation between each stage. For that reason, I've assigned to a temporary variable in the "act" stage to separate this from the "assert" stage.

Your point is very valid. I've chosen to keep these separate for consistency with the other tests but I've no objections to changing this.

harryBedfordSL (approved):

PR #44, opened 2024-05-01
separate code into src and tests directories

harryBedfordSL:
## Description

This PR separates out our test code from our source code. By doing this our docker file is slightly more optimised, as we are now only copying over the functional code to be part of the container; the tests are not needed.

## Changelog
- create new root level `src` directory
- create new root level `tests` directory
- Move all backend code into respective places
- Update the backend docker file

## Notes
- Running the backend tests remains unchanged: `pytest`
- I have verified that `docker compose up` is still working
- Now, to run the backend directly (from `/backend`) the uvicorn command changes to: `uvicorn src.api:app --port 8250`


backend/README.md
```diff
@@ -69,13 +69,16 @@ To add further rules, these are added to `ruff.toml` by using the letter asssign
```
Ch0live, line 69: Line 25 needs updating to
```
uvicorn src.api:app --port 8250
```

harryBedfordSL: great spot!

harryBedfordSL: updated 🙂

Ch0live:
- Local run works
- Docker run works
- All (33) tests pass

Great stuff!

backend/src/utils/__init__.py
```diff
@@ -1,12 +1,13 @@
 from .config import Config
-from .llm import call_model, call_model_with_tools, get_response_three_prompts
+from .llm import call_model, call_model_with_tools, get_response_three_prompts, MistralClient
```
Ch0live, line 2: Do we actually use `MistralClient` outside of the `/utils` directory?

Ch0live (changes requested):
1 other comment about MistralClient added

backend/src/utils/__init__.py
harryBedfordSL: we use it in the tests

Ch0live (dismissed):

bheyman-scottlogic:
I think that under the `/src` directory it should only be directories for ease of navigation and that shouldn't necessarily be a package with an `__init__.py` . Happy to discuss what you think is more appropriate?

Ch0live (approved):

bheyman-scottlogic (approved):
Looks good, nice work Harry!

PR #46, opened 2024-05-01
Loop over agents until task is resolved

Ch0live:
## Description

First attempt at creating the complete Router that can iterate over multiple tasks. WIP

## Changelog
-


backend/src/agents/goal_achieved_agent.py
```diff
@@ -0,0 +1,10 @@
+from .agent import Agent, agent_metadata
+
+@agent_metadata(
+    name="GoalAchievedAgent",
+    prompt="",
+    description="Pick this agent if you believe you have solved BOTH the Current Task and the Next Task",
```
hsauve-scottlogic, line 6: Just a detail but the description for other agents is about what the agent does 'this agent does this' rather than an instruction to another agent, so to follow the same pattern we could have something like 'This agent informs the user that the goal has been achieved'

backend/src/agents/unresolvable_task_agent.py
```diff
@@ -0,0 +1,10 @@
+from .agent import Agent, agent_metadata
+
+@agent_metadata(
+    name="UnresolvableTaskAgent",
+    description="Pick this agent if you believe (based on the History of actions) you cannot solve the Current Task.",
```
hsauve-scottlogic, line 5: same comment as above

backend/src/supervisors/supervisor.py
```diff
@@ -1,30 +1,86 @@
...
+    while still_solving_tasks(current_agent_name, attempts_count):
+        logging.info(f"attempts_count: {attempts_count}, task_step: {task_step}")
+        logging.info(f"current_agent: {current_agent_name}, history: {history}")
+
+        # Assign tasks
+        current_task = str(tasks_json["tasks"][task_step])
+
+        # If on the last task don't iterate
+        if (task_step == task_total_count-1):
+            next_task = "There is no next task - the Next Task is the final task (sorry I know this is confusing)"
```
hsauve-scottlogic, line 45: Sorry Charlie, who is this message to? The LLM?

backend/tests/prompts/prompting_test.py
```diff
@@ -1,64 +1,229 @@
...
+AGENT LIST:
+
+
+If none of the agents are appropriate for solving this goal, choose "UnresolvableTaskAgent".
+If you believe you have solved the final problem, choose "GoalAchievedAgent"
+
+
+## Determine the next best step
+Your task is to pick one of the mentioned agents above to complete the task.
+If you\'re last attempt at the Current Task looks successful, move on to the Next Task
```
hsauve-scottlogic, line 129:
```suggestion
If your last attempt at the Current Task looks successful, move on to the Next Task
```

hsauve-scottlogic (commented):
Looks really Charlie, nice work! Just a few comments

backend/src/supervisors/supervisor.py
```diff
@@ -1,30 +1,86 @@
...
+
+
+def still_solving_tasks(current_agent, iter_count):
+    agent_calling_unresolved = current_agent != "UnresolvableTaskAgent" and current_agent != "GoalAchievedAgent"
+    count_not_exceeded = iter_count < 5
+    return agent_calling_unresolved and count_not_exceeded
+
+
+def onto_next_task(best_next_step_json):
+    return best_next_step_json["current_or_next_task"].__contains__("next")
```
DannyNicholas, line 20: From what I've read, we shouldn't be calling "magic" name methods directly like `___contains__`.

If I understand the flow correctly, this is actually a Python dictionary (??) so we can use:
```
return "next" in best_next_step_json["current_or_next_task"]
```

backend/src/supervisors/supervisor.py
```diff
@@ -1,30 +1,86 @@
...
+        best_next_step_json = pick_agent(current_task, next_task, list_of_agents, history)
+        current_agent_name = best_next_step_json["agent"]
+
+        # Check if we are done
+        if current_agent_name == "UnresolvableTaskAgent":
+            logging.info("UnresolvableTaskAgent called :( returning fail case")
+            return "I am sorry, but I was unable to find an answer to your question"
+        if current_agent_name == "GoalAchievedAgent":
+            logging.info("goal achieved!")
+            return best_next_step_json["thoughts"]["speak"] # TODO: Properly return answer when problem is solved
```
DannyNicholas, line 59: Feels like this logic should be encapsulated within each specific agent. The supervisor shouldn't have knowledge of how each agent behaves.

DannyNicholas (dismissed):
Looks good @Ch0live - raised a number of comments and questions. Happy to go through this together on a call if that's helpful?

One thing I've found confusing is that we appear to be converting to/from JSON quite often so at various stages it's not clear if we're handling python dictionaries or strings.

Would it be possible to keep data as python dictionaries as long as possible to simplify the code?

backend/src/agents/goal_achieved_agent.py
Ch0live: It doesn't directly return to the user the result of the process, more that it acts as a choice for the router that, when chosen, has the side-affect that the process terminates. I do agree it can be named more in line with the other agents however.

Have changed to `"This agent triggers the end of solving the overall problem and returns the final answer to the user"`

backend/src/agents/unresolvable_task_agent.py
Ch0live: Also updated

backend/tests/prompts/prompting_test.py
Ch0live: Updated

backend/src/supervisors/supervisor.py
Ch0live: It was a note I left in haste when frantically creating this bit of logic, hence the humor. The message is for the LLM to explicitly tell it to just solve the current task and not the next task.

I have refactored this to be clearer

backend/src/agents/goal_achieved_agent.py
hsauve-scottlogic: Sounds good and makes sense, thanks Charlie

backend/tests/prompts/prompting_test.py
```diff
@@ -19,9 +19,9 @@ def test_load_agent_selection_format_template():
         \\"plan\\": \\"description of the plan for the chosen agent\\",
         \\"reasoning\\": \\"reasoning behind choosing the agent\\",
         \\"criticism\\": \\"constructive self-criticism\\",
-        \\"speak\\": \\"thoughts summary to say to user on 1. if you're solving the current or next task and why 2. which agent you've chosen and why\\",
+        \\"speak\\": \\"thoughts summary to say to user on 1. if your solving the current or next task and why 2. which agent you've chosen and why\\",
     },
-    \\"current_or_next_task\\": \\"either the word "current" or the word "next" depending on which task you're trying to solve\\",
+    \\"current_or_next_task\\": \\"either the word "current" or the word "next" depending on which task your trying to solve\\",
```
DannyNicholas, line 24: I think `you're` is grammatically correct here (and above).

backend/src/agents/goal_achieved_agent.py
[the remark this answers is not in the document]
Ch0live: So the idea is that 
- if we have done nothing so far, then the current and next task will be considered unresolved and we will attempt to resolve the current task
- if we have resolved the current task, then we look to solving the next task
Note that if we are on the final task we set the next task to "There is no next task. Please solve the Current Task". Therefore
- if we have solved all tasks, we should pick this agent. And that scenario is when we believe to have solved both the current and next tasks

We effectively check if all tasks are solved by being on the last few tasks and checking if just those are solved.

frontend/src/server.ts
```diff
@@ -29,6 +29,7 @@ const checkBackendHealth = async (): Promise<ChatMessageResponse> => {
 const unhappyChatResponse = createChatMessageResponse('I\'m sorry, but I was unable to process your message. Please check the status of the service using the phrase "healthcheck"');
 
 const callChatEndpoint = async (message: string): Promise<ChatMessageResponse> => {
+  console.log(process.env.BACKEND_URL)
```
DannyNicholas, line 32: missing semi-colon. Linter is sad 😢

backend/src/agents/goal_achieved_agent.py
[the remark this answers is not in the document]
DannyNicholas: Thanks for clarifying.

backend/src/agents/datastore_agent.py
```diff
@@ -0,0 +1,41 @@
...
+        )
+    },
+)
+def get_month_data(merchant: str) -> str:
+    if merchant.__contains__("Amazon"):
+        return "Last month you spend £64.21 on Amazon"
+    elif merchant.__contains__("Netflix"):
+        return "Last month you spend £6.99 on Netflix"
+    else:
+        return "Last month you didn't spend anything on " + merchant
```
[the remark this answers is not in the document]
Ch0live: I agree. Updated as you have suggested here.

backend/src/agents/agent.py
```diff
@@ -25,6 +26,8 @@ def __get_action(self, utterance: str) -> Action_and_args:
 
     def invoke(self, utterance: str) -> str:
         (action, args) = self.__get_action(utterance)
+        result_of_action = action(**args)
+        logging.info(f"Tool \"{action.name}\" chosen for agent \"{self.name}\" with result: {result_of_action}")
         return action(**args)
```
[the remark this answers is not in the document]
Ch0live: Good spot, changed

backend/src/agents/maths_agent.py
```diff
@@ -0,0 +1,47 @@
...
+            type="number",
+            description="value of first thing",
+        )
+    },
+)
+def compare_two_values(value_one, thing_one, value_two, thing_two) -> str:
+    if (value_one > value_two):
+        return f"You have spent more on {thing_one} ({value_one}) than {thing_two} ({value_two}) in the last month"
+    else:
+        return f"You have spent more on {thing_two} ({value_two}) than {thing_one} ({value_one}) in the last month"
```
DannyNicholas, line 32: can we drop `in the last month`? (here and line above)

I believe this code is meant to be generic so can compare two values over any time period.

backend/src/supervisors/supervisor.py
```diff
@@ -1,30 +1,86 @@
...
+goal_achieved_agent = GoalAchievedAgent()
+list_of_agents = [ DatastoreAgent(), MathsAgent(), goal_achieved_agent, unresolvable_task_agent ]
+
+prompt_engine = PromptEngine()
+
+
+def still_solving_tasks(current_agent, iter_count):
+    agent_calling_unresolved = current_agent != "UnresolvableTaskAgent" and current_agent != "GoalAchievedAgent"
+    count_not_exceeded = iter_count < 5
+    return agent_calling_unresolved and count_not_exceeded
```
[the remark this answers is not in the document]
Ch0live: **If we want the agents to be responsible to move on to the next task** we may run the risk of calling the wrong agent that produces a convincing answer. What if we wanted to ask about the users favorite sports team, but instead the router mistakenly chose to search the web and found the generally most popular team? The agent for searching the web would probably be convinced the issue is solved and that we can move on. If we don't verify with the router post completion that the _correct problem_ has been solved via picking the correct agent I think we always run this risk.

**If we want these agents to handle terminating the process**, I'm not sure how this would work. The while loop requires breaking for us to quit the flow either from solving the question (utterance) or abandoning it. If the agent contains the logic it will need to be able to (within it's `invoke()` method) break that loop and return the final value, and I'm not sure of a clean way to put that logic within the agent instead of as a check in the `supervisor.py`

backend/src/supervisors/supervisor.py
Ch0live: This logic is only in the case that _all_ tasks have been solved. No agent should be doing this return statement, only if we have completed the flow should we be concerned with it.

backend/tests/supervisors/supervisor_test.py
```diff
@@ -13,11 +10,9 @@
...
-    try:
-        solve_all_tasks(tasks_json)
-    except Exception:
-        raise
+# TODO: Create test that mocks out LLM call
+# def test_load_task_step_template():
+#     try:
+#         solve_all_tasks(json.loads(hair_task))
+#     except Exception:
+#         raise
```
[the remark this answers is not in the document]
Ch0live: For now I will delete it, again the work is captured in new tickets

backend/src/router.py
```diff
@@ -1,35 +1,52 @@
...
 prompt_engine = PromptEngine()
 
 
-def pick_agent(task_string):
-    logging.debug("Picking agent for task: " + task_string)
+def convert_agents_to_json_shaped_string(agent):
+    agent_as_object = {
+        "name": agent.name,
+        "description": agent.description
+    }
```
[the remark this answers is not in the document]
Ch0live: Updated to `create_agent_object`

backend/src/router.py
```diff
@@ -1,35 +1,52 @@
...
-    list_of_agents = ["unresolvable_agent, database_agent, financial_advisor_agent, web_search_agent"]
 
-    agent_list_prompt = prompt_engine.load_prompt("agents-list", list_of_agents=list_of_agents)
-    response_format_prompt = prompt_engine.load_prompt("agent-selection-format")
-    best_next_step_prompt = prompt_engine.load_prompt("task-step", task=task_string)
+def convert_step_to_json(next_step):
+    try:
+        return json.loads(next_step)
+    except Exception:
+        raise Exception("Failed to interpret LLM next step format")
```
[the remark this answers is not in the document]
Ch0live: Yes, have added to log

backend/src/router.py
```diff
@@ -1,35 +1,52 @@
...
-    # collect all required prompts
-    # system-prompt: List of agents that can be called
-    # system-prompt: Expected response shape (probably a json)
-    # user-prompt: "pick an agent and fit to the response shape to solve {taskString}
+# TODO: Create pick_agent test with mocked calls
+def pick_agent(current_task_string, next_task_string, list_of_agents, history):
+    logging.debug("Picking agent")
 
-    logging.debug("Found next best step:")
+    list_of_agent_string = [convert_agents_to_json_shaped_string(agent) for agent in list_of_agents]
```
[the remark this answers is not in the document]
Ch0live: Updated. `json.dumps()` converts objects to strings with nice formatting so have updated the current and next task names too.

backend/src/supervisors/supervisor.py
Ch0live: This is simply checking if the router has chosen to solve the current or next task, and we're returning True if we're completing the next one and False if it's the previous.

I can see using "magic" names being bad, have adjusted to just being equivalent (from manual testing the router is quite good at returning either "current" or "next" with no variation from those strings)

backend/tests/prompts/prompting_test.py
Ch0live: O yeah 🤦 reverted

frontend/src/server.ts
Ch0live: Ah! Also not meant to be there. Removed

backend/src/agents/maths_agent.py
Ch0live: We can, but it is also still very much a stub response and should be treated as one. I'm not a huge fan of half converting this method into a working comparison method. Would rather refactor the `MathsAgent` all in 1 go to no longer be a stub than tweak this part

DannyNicholas: Nor me.... sorry I didn't realise we still need this to be a stub.

I assumed converting this to a real implementation would be fine as it's using values being passed-in. I assumed only the `datastore_agent` needed stubs as we're only missing the database.

If you want to roll this back to be a 100% stub, please do.

DannyNicholas (approved):
Thanks for all the changes @Ch0live  - I've approved to avoid holding anything up as I'm leave tomorrow.

I still think the we need to re-visit how the supervisor uses agents but that can wait until we review this together and can be addressed in a future PR.

hsauve-scottlogic (approved):

Ch0live:
I added any hanging comments onto the comments of the relevant issues to be refined further

PR #47, opened 2024-05-01
Create pytest GitHub action

DannyNicholas:
## Description

Configure a GitHub action to run `pytest` on our backend code. This action will run following every update to a pull-request. GitHub will be configured so that all tests must pass before we can merge.

Sorry for the **many** commits. I was testing a few options to run the tests.

---

## Test Failures

After implementing this, two tests were shown to fail:

- `backend/agents/agent_test.py` = `test_agent_invoke_uses_tool`
- `backend/supervisors/supervisor_test.py` = `test_load_task_step_template`

They were making real calls to our LLM (that will not available from within a GitHub runner). The failures were confirmed locally by running `pytest` on my laptop while disconnected from the network.

These tests have been fixed as part of this PR.

---

## Reporting

The `pytest` action compiles results into a summary report. Example output of tests:

**NOTE:** This shows test failures (prior to the tests being fixed)

image

---

## Changelog

- Add a `test-backend.yml` file to trigger a GitHub action to run `pytest` on our backend code.


.github/workflows/test-backend.yml
```diff
@@ -0,0 +1,47 @@
+name: Test Backend
+on: 
+  pull_request:
```
harryBedfordSL, line 3: was it part of this ticket to look into the `path` option here?

.github/workflows/test-backend.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          pip install pytest-md pytest-emoji
```
harryBedfordSL, line 25: should these two be in our requirements file, or are they github action only requirements?

.github/workflows/test-backend.yml
```diff
@@ -0,0 +1,47 @@
...
+
+        # some tests require an .env file to run.
+        # file should contain all the required properties.
```
harryBedfordSL, line 28: oh interesting, should we be mocking the config for all tests so wouldn't need this?

.github/workflows/test-backend.yml
```diff
@@ -0,0 +1,47 @@
...
+          NEO4J_USERNAME=neo4j
+          NEO4J_PASSWORD=my-password
+          EOF
+          cat .env
+
+      - name: Run tests
+        uses: pavelzw/pytest-action@v2
+        with:
+          emoji: true
+          verbose: true
```
harryBedfordSL, line 45: ah I'm guessing this is where those requirements are needed? Is this action better than just running `pytest` without a marketplace action?

harryBedfordSL (commented):
looks great, just some questions from me

.github/workflows/test-backend.yml
DannyNicholas: these are only needed by the `pytest-action@v2` GitHub action to create a report in markdown and add emojis to the report.

.github/workflows/test-backend.yml
DannyNicholas: probably 😄 

Without the environment file, 7 tests fail. One example is the neo4j driver throwing errors as it needed a valid URL to initialise - although the tests don't seem to actually use the driver.

Since this could require a bit of refactoring, would you be happy if I investigated/fixed in a separate PR?

.github/workflows/test-backend.yml
DannyNicholas: The major benefit is the reporting. You get a nice summary in markdown. I'll give you a demo

.github/workflows/test-backend.yml
DannyNicholas: ermm..... yes it was 😃 

Updated with a path filter so it only runs for changes with the `backend` folder.

.github/workflows/test-backend.yml
harryBedfordSL: yeah of course, I reckon a separate ticket to investigate / refactor sounds great

.github/workflows/test-backend.yml
harryBedfordSL: Thanks for the demo, the reports look really cool

harryBedfordSL (dismissed):

Ch0live (approved):
Looks good - the need for stub config is my only comment but I don't think it's something that needs resolving here

PR #48, opened 2024-05-02
add instructions on how to turn on type checking for backend

harryBedfordSL:
## Description

Small PR to add instructions to the backend readme on how to enable basic type checking for our python code. This helps developers see errors when types are not correct 

## Changelog
- Updated backend readme with contributing instructions

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -27,7 +27,7 @@ uvicorn src.api:app --port 8250
 
 4. Check the backend app is running at [http://localhost:8250/health](http://localhost:8250/health)
 
-## Backend Linting
+## Linting
 
 Ruff is being used for the backend linting. See the docs: [here](https://docs.astral.sh/ruff/)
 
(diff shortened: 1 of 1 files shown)


hsauve-scottlogic (approved):
Looks great Harry, instructions are clear :)

backend/README.md
```diff
@@ -82,3 +82,17 @@ pytest
...
 More documentation on `pytest` can be found [here](https://docs.pytest.org/en/8.0.x/).
 
+## Contributing
+
+### Type errors
+
+The vscode extension [Pylance](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance) can be used to highlight basic type errors in your python code. To set this up do the following:
+
+- Install the extension
+- Add the following to your `../.vscode/settings.json` file:
```
philbaxter, line 92: A bit ambiguous, my `.vscode` directory is in my user home directory, specifying `..` doesn't get me there on my system, maybe `$HOME/.vscode/settings.json` would be a better pointer.

philbaxter (changes requested):

backend/README.md
harryBedfordSL: you can have two `settings.json` , your user one as you say is at `$HOME/.vscode` , but you can also have a workspace one, which just applies to the current project , which lives under the `<project_root>/.vscode` . I kept it as the workspace one here as thought is more relavant, but maybe I should mention both options? What do you think?

philbaxter: Ah, fair enough, I only have the one in $HOME/.vscode. I didnt realise that you could have a workspace one, makes sense though.

PR #54, opened 2024-05-13
Refactor supervisor router

harryBedfordSL:
## Description

This PR refactors the supervisor and router, along with adding tests for each. The main changes are flattening of the looping logic. We no longer use an agent as the trigger to move on to the next task, but instead have a single `validator_agent` . The supervisor calls on this agent with each task answer to see if it should move on to the next task, or retry the current one. 

Another notable change is how we are doing the loop within the supervisor. Instead of a while loop we are now using recursion to keep attempting to solve the current task, still up to a maximum of 5 attempts.

## Changelog
- supervisor `solve_all_tasks` simplified
- New `validator_agent` instead of `unresolvable_agent` and `goal_achieved_agent`
- router renaming and minor splitting of methods
- new `to_json` utility method added as noticed we were using it in a couple of places
- centralised a `MockAgent` in `tests.Agents` for reusability across tests

## Testing

I was using the `What did I spend more on last month, Amazon or Netflix?` question to make sure that the same expected response was being seen:

image


backend/src/agents/__init__.py
```diff
@@ -3,8 +3,26 @@
...
-from .unresolvable_task_agent import UnresolvableTaskAgent
+from .validator_agent import ValidatorAgent
 
-__all__ = ["Agent", "DatastoreAgent", "MathsAgent", "GoalAchievedAgent", "UnresolvableTaskAgent", "agent_metadata",
-           "create_tasks", "tool_metadata", "Parameter", "convert_to_mistral_tool"]
+validator_agent = ValidatorAgent()
+
+
+def get_agent_details(agent):
+    return {"name": agent.name, "description": agent.description}
```
Ch0live, line 12: Neat idea having a getter method here 🚀

backend/src/agents/__init__.py
```diff
@@ -3,8 +3,26 @@
...
+
+
+agents = [DatastoreAgent(), MathsAgent()]
+agents_details = [get_agent_details(agent) for agent in agents]
```
Ch0live, line 16: Neat idea collecting all the agent details here like this 👍

backend/src/agents/validator_agent.py
```diff
@@ -0,0 +1,42 @@
+from src.utils import call_model
+from src.agents import Agent, agent_metadata
+
+validator_prompt = """
```
Ch0live, line 4: Are we adding all prompts as jinja2 templates? I understand it doesn't take any parameters so would be imported just as  a large string but would potentially be cleaner than having it here as a string

backend/src/agents/validator_agent.py
```diff
@@ -0,0 +1,42 @@
...
+Response: False
+
+Task: Find all spending transactions last month on Amazon.
+Answer: Last month you spend £64.21 on Amazon
+Response: True
+
+Task: Find all spending transactions last month on Amazon.
+Answer: Last month you spend £64.21 on Spotify
+Response: False
+Reasoning: The answer is for Spotify not Amazon.
```
Ch0live, line 27:
```
Reasoning: The answer is for Spotify not Amazon.
```

Do we need this line? It might confuse the LLM to return a response with a Reasoning field. Saying this with no example of it doing so but wanted to raise the concern all the same - if it performs as well without I personally think we should remove it

backend/src/prompts/templates/best-next-step.j2
```diff
@@ -30,13 +22,9 @@ Here is the list of Agents you can choose from:
...
 {{ list_of_agents }}
 
-If none of the agents are appropriate for solving this goal, choose "UnresolvableTaskAgent".
-If you believe you have solved the final problem, choose "GoalAchievedAgent"
-
-
 ## Determine the next best step
 Your task is to pick one of the mentioned agents above to complete the task.
-If you're last attempt at the Current Task looks successful, move on to the Next Task
+If from the history you see that the agent has failed to complete a task (by repeating a task with the same agent), do not assign the task to them again.
```
Ch0live, line 27: Good touch to the prompt 🚀

backend/src/router.py
```diff
@@ -1,42 +1,27 @@
...
-    best_next_step_prompt = prompt_engine.load_prompt(
-        "best-next-step",
-        current_task=json.dumps(current_task_object, indent=4),
-        next_task=json.dumps(next_task_object, indent=4),
-        list_of_agents=json.dumps(available_agents_object, indent=4),
-        history=history
-    )
 
-    response_format_prompt = prompt_engine.load_prompt("agent-selection-format")
+def get_plan(task, scratchpad):
```
Ch0live, line 23: Nit-picky, but as we're creating an original plan and not retrieving one, could this be called `generate_plan` or `create_plan`?

Ch0live: Maybe `build` like we use above in `build_best_next_step_prompt`?

backend/tests/prompts/prompting_test.py
```diff
@@ -63,18 +58,14 @@ def test_load_best_next_step_template():
...
 
 
-If none of the agents are appropriate for solving this goal, choose "UnresolvableTaskAgent".
-If you believe you have solved the final problem, choose "GoalAchievedAgent"
-
-
 ## Determine the next best step
 Your task is to pick one of the mentioned agents above to complete the task.
-If you\'re last attempt at the Current Task looks successful, move on to the Next Task
+If from the history you see that the agent has failed to complete a task (by repeating a task with the same agent), do not assign the task to them again.
```
Ch0live, line 63: This part of the prompt may need expanding or highlighting further. Or even perhaps we should make choosing the same agent a hard-coded blocker (remove it from the list of agents to choose from the second time for example).

My first run of the example prompt had the Netflix transactions call MathsAgent 5 times
```
INFO:     Final scratchpad:
INFO:     [
    {
        "agent_name": "DatastoreAgent",
        "task": "Find all transactions on Netflix from the user's account last month.",
        "result": "Last month you spend \u00a36.99 on Netflix"
    },
    {
        "agent_name": "DatastoreAgent",
        "task": "Find all transactions on Amazon from the user's account last month.",
        "result": "Last month you spend \u00a364.21 on Amazon"
    },
    {
        "agent_name": "MathsAgent",
        "task": "Sum up the transactions for Netflix.",
        "result": "You have spent more on Netflix transaction 1 (15.99) than Netflix transaction 2 (13.99) in the last month"
    },
    {
        "agent_name": "MathsAgent",
        "task": "Sum up the transactions for Netflix.",
        "result": "You have spent more on Netflix transaction 1 (15.99) than Netflix transaction 2 (13.99) in the last month"
    },
    {
        "agent_name": "MathsAgent",
        "task": "Sum up the transactions for Netflix.",
        "result": "You have spent more on Netflix transaction 1 (15.99) than Netflix transaction 2 (13.99) in the last month"
    },
    {
        "agent_name": "MathsAgent",
        "task": "Sum up the transactions for Netflix.",
        "result": "You have spent more on Netflix transaction 1 (15.99) than Netflix transaction 2 (13.99) in the last month"
    },
    {
        "agent_name": "MathsAgent",
        "task": "Sum up the transactions for Netflix.",
        "result": "You have spent more on Netflix transaction 1 (15.99) than Netflix transaction 2 (13.99) in the last month"
    },
    {
        "agent_name": null,
        "task": "Sum up the transactions for Netflix.",
        "result": "I am sorry, but I was unable to find an answer to this task"
    },
    {
        "agent_name": null,
        "task": "Sum up the transactions for Amazon.",
        "result": "I am sorry, but I was unable to find an agent to solve this task"
    },
    {
        "agent_name": "MathsAgent",
        "task": "Compare the total spending on Netflix and Amazon.",
        "result": "You have spent more on Netflix (1500) than Amazon (1200) in the last month"
    }
]
```

If this is an anomaly or not I'm not sure, but it seems it will pick the same agent multiple times without any other good choices

Ch0live (changes requested):
Great refactor 💯, a few comments with some requested changes

backend/src/agents/validator_agent.py
harryBedfordSL: a very good point. Yeah I'll move it to a jinga template, I agree it would be neater

backend/src/agents/validator_agent.py
harryBedfordSL: hmm I think it helped whilst I was testing. I think the next couple of lines are enough to make sure it only returns the boolean

backend/src/agents/validator_agent.py
```diff
@@ -0,0 +1,42 @@
...
+
+@agent_metadata(
+    name="ValidatorAgent",
+    description="This agent is responsible for validating the answers to the tasks",
+    prompt=validator_prompt,
+    tools=[],
+)
+class ValidatorAgent(Agent):
+    def invoke(self, utterance: str) -> str:
+        return call_model(self.prompt, utterance)
```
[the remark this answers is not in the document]
harryBedfordSL: yeah happy to move the log

backend/src/router.py
harryBedfordSL: yeah good point, like the `build` suggestion for consistency

backend/tests/prompts/prompting_test.py
harryBedfordSL: I'll try a tweaked version

backend/src/agents/__init__.py
```diff
@@ -3,8 +3,26 @@
...
-__all__ = ["Agent", "DatastoreAgent", "MathsAgent", "GoalAchievedAgent", "UnresolvableTaskAgent", "agent_metadata",
-           "create_tasks", "tool_metadata", "Parameter", "convert_to_mistral_tool"]
+validator_agent = ValidatorAgent()
+
+
+def get_agent_details(agent):
+    return {"name": agent.name, "description": agent.description}
+
+
+agents = [DatastoreAgent(), MathsAgent()]
```
hsauve-scottlogic, line 15: Why are some agents listed as agents, but create_tasks for example isn't in there?

backend/src/prompts/templates/validator.j2
```diff
@@ -0,0 +1,26 @@
+You are an expert validator. You can help with validating the answers to the tasks with just the information provided.
+
+You're entire purpose is to return a boolean value to indicate if the answer has fulfilled the task.
```
hsauve-scottlogic, line 3:
```suggestion
Your entire purpose is to return a boolean value to indicate if the answer has fulfilled the task.

```

backend/tests/prompts/prompting_test.py
```diff
@@ -120,22 +101,14 @@ def test_load_best_next_step_with_history_template():
...
 
 
-If none of the agents are appropriate for solving this goal, choose "UnresolvableTaskAgent".
-If you believe you have solved the final problem, choose "GoalAchievedAgent"
-
-
 ## Determine the next best step
 Your task is to pick one of the mentioned agents above to complete the task.
-If you\'re last attempt at the Current Task looks successful, move on to the Next Task
+If the same agent_name and task are repeated more than twice in the histroy, you must not pick that agent_name.
```
hsauve-scottlogic, line 106:
```suggestion
If the same agent_name and task are repeated more than twice in the history, you must not pick that agent_name.
```

hsauve-scottlogic (commented):
Looks really good Harry, nice tidying up. Just a couple of questions

backend/src/router.py
```diff
@@ -1,42 +1,27 @@
...
+        list_of_agents=json.dumps(agents_details, indent=4),
+        history=json.dumps(scratchpad, indent=4),
+    )
 
-# TODO: Create pick_agent test with mocked calls
-def pick_agent(current_task_object, next_task_object, agents_list, history):
-    logging.debug("Picking agent")
 
-    available_agents_object = [create_agent_object(agent) for agent in agents_list]
+response_format_prompt = prompt_engine.load_prompt("agent-selection-format")
```
Ch0live, line 20: Is this intended correctly?

Ch0live (approved):

backend/src/router.py
harryBedfordSL: yes, I lifted the prompt out of the method as is static / won't change

backend/src/agents/__init__.py
harryBedfordSL: Create tasks agent doesn't currently use our class / decorators, and this is the list of agents that are available to the supervisor / router, so create tasks probably shouldn't be in this array, unless we want that ability in the future 🤔

backend/src/prompts/templates/validator.j2
harryBedfordSL: great spot

Ch0live (approved):

backend/src/agents/__init__.py
hsauve-scottlogic: Aah yes I see, thanks I understand the logic now :)

PR #55, opened 2024-05-14
Add custom function calling

Ch0live:
## Description

We are no longer bound to Mistral function calling! This PR adds custom function calling logic through `best-tool.j2` and `tool-selection-format.j2` prompts to select a tool using the regular chat endpoint that all LLM APIs offer.

image

## Changelog
- Removed `convert_to_mistral_tool` and `call_model_with_tools(system_prompt, user_prompt, tools)`
- Added `"##### ~ Milestone logs ~ #####"` to key points in the flow to improve log readability
- Added `sum_list_of_values(list_of_values)` tool to `MathsAgent`
- Remove the need for agent prompts (they were only needed for the tool selection prompt which works independent of the type of agent
- Added new prompts for tool selection `best-tool.j2` and `tool-selection-format.j2`
- Added `scratchpad.py` to make getting and setting the scratchpad easier (removes circular imports as well)
- Remove unnecessary `\"` in prompts (instead just using `"`)


backend/src/agents/adapters.py
```diff
@@ -1,22 +1,54 @@
...
+def to_object(tool: Tool) -> str:
+    obj = {
+        "description": tool.description,
+        "name": tool.name,
+        "parameters": {
+            key: {
+                    "type": inner_dict.type,
+                    "description": inner_dict.description,
+                }
+                for key, inner_dict in tool.parameters.items()
```
harryBedfordSL, line 19: Should this method live on every tool class?

backend/src/agents/agent.py
```diff
@@ -1,41 +1,58 @@
...
     description: str
     tools: List[Tool]
-    prompt: str
-
     def __get_action(self, utterance: str) -> Action_and_args:
-        tools = map(convert_to_mistral_tool, self.tools)
-        (function_name, function_params) = call_model_with_tools(self.prompt, utterance, tools)
 
-        tool = next((tool for tool in self.tools if tool.action.__name__ == function_name), None)
+        format_prompt = engine.load_prompt("tool-selection-format")
```
harryBedfordSL, line 20: could live outside of this class, right below the engine initialisation on line 12, to avoid being re created on every __get_action call

backend/src/utils/scratchpad.py
```diff
@@ -0,0 +1,13 @@
...
+def get_scratchpad() -> list:
+    return scratchpad
+
+
+def update_scratchpad(agent_name, task, result):
+    scratchpad.append({"agent_name": agent_name, "task": task["summary"], "result": result})
+
+
+def clear_scratchpad():
+    scratchpad.clear()
```
harryBedfordSL, line 13: nice! Python question from me, will only one scratchpad instance ever be created, regardless of how many things import this file?

backend/src/agents/adapters.py
Ch0live: I'm confused what you mean. The `to_object()` method is exclusively used in `create_all_tools_str()` and that is a dependency on the `agent.py` invoke method itself.

Are you suggesting we move it to the `tool.py` file? I don't believe we have a tool class per se but if we did it would be a good place to house this

backend/src/agents/agent.py
Ch0live: Yes, agreed.

Moved outside the class

backend/src/utils/scratchpad.py
Ch0live: We don't import the variable, just the methods. Therefore, I believe all method calls with access the same scratchpad, regardless of where they're imported to. So I'm gonna say yes, but only because I don't see how we would be "newing up" a new scratchpad variable without importing the variable itself anywhere

backend/src/agents/adapters.py
harryBedfordSL: Yeah the latter, I mean should / can it live on this class ? https://github.com/WaitThatShouldntWork/InferGPT/blob/8e135eb5964700ecf9558b60b6dc4e20e5bbec17/backend/src/agents/types.py#L20

backend/src/prompts/templates/best-tool.j2
```diff
@@ -0,0 +1,25 @@
...
+
+below is your history of all work you have assigned and had completed by Agents
+Trust the information below completely (100% accurate)
+
+{{ scratchpad }}
+
+Pick 1 tool (no more than 1) from the list below to complete this task.
+Fit the correct parameters from the task to the tool arguments.
+If the parameters are tagged as optional, you do not need to fill them in,
+but feel free to if it is necessary
```
bheyman-scottlogic, line 15: I feel this wording can be improved to say that although optional you should try and fill in but do not make something up to fill it. (something along those lines)

backend/src/prompts/templates/validator.j2
```diff
@@ -4,6 +4,8 @@ Your entire purpose is to return a boolean value to indicate if the answer has f
 
 You will be passed a task and an answer. You need to determine if the answer is correct or not.
 
+Be lenient - if the answer looks reasonably right (isn't a million miles off) then return True
```
bheyman-scottlogic, line 7: I think the phrase "isn't a million miles off" is a bit misleading for the LLM

backend/src/utils/scratchpad.py
harryBedfordSL: sounds good to me. Yeah my only concern was if the whole script `scratchpad.py` is run every time we import one of the methods, but sounds like that is not the case

bheyman-scottlogic (commented):
Nice work Charlie, just added a couple of comments

backend/src/agents/adapters.py
```diff
@@ -1,22 +1,54 @@
...
-                    key: {
-                        "type": inner_dict.type,
-                        "description": inner_dict.description,
-                    }
-                    for key, inner_dict in tool.parameters.items()
-                },
-                "required": [key for key, inner_dict in tool.parameters.items() if inner_dict.required],
-            },
+def to_object(tool: Tool) -> str:
+    obj = {
```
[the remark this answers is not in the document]
Ch0live: Hows `convert_tool_to_obj_str` as it's actually returning a string type?

backend/src/agents/adapters.py
```diff
@@ -1,22 +1,54 @@
...
+    if chosen_tool_name == "None":
+        raise Exception("No tool deemed appropriate for task")
+    try:
+        tool = next(tool for tool in agent_tools if tool.name == chosen_tool_name)
+    except Exception:
+        raise Exception(f"Unable to find tool {chosen_tool_name} in available tools")
+    return tool
+
+
+def get_required_args(tool: Tool):
```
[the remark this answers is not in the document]
Ch0live: Good spot, changed 👍

backend/src/prompts/templates/best-tool.j2
Ch0live: Updated

backend/src/prompts/templates/validator.j2
Ch0live: Yeah in hindsight I totally agree, removed

backend/src/agents/adapters.py
Ch0live: I have made these changes as discussed on call in [e5082d7](https://github.com/WaitThatShouldntWork/InferGPT/pull/55/commits/e5082d744aaedd9407e8a7e0d9c0defd61853e1e)

harryBedfordSL (approved):
Looks great to me Charlie, nice one!

backend/tests/agents/tool_test.py
```diff
@@ -50,4 +49,9 @@ def test_tool_optional_required():
 
 
 def test_tool_action():
-    assert mock_tool.action == mock_tool
+    assert mock_tool.action() == "Hello, World!"
+
+
+expected_tools_object = """{"description": "A test tool", "name": "Mock Tool", "parameters": {"input": {"type": "string", "description": "A string"}, "optional": {"type": "string", "description": "A string"}}}""" # noqa: E501
+def test_to_object():
```
bheyman-scottlogic, line 56: Whitespace before function

bheyman-scottlogic (approved):
Just some minor formatting points, good work Charlie!

backend/src/agents/tool.py
```diff
@@ -1,13 +1,33 @@
...
+                    }
+                    for key, inner_dict in self.parameters.items()
+            },
+        }
 
-    return decorator
+        return json.dumps(obj)
+
+
+def tool(name: str, description: str, parameters: dict[str, Parameter]):
```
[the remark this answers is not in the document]
Ch0live: Adjusted in a new branch [here](https://github.com/WaitThatShouldntWork/InferGPT/tree/feature/refactor-director)

backend/src/agents/tool.py
```diff
@@ -1,13 +1,33 @@
...
-        tool.action = tool
+class Tool():
+    def __init__(self, name: str, description: str, parameters: dict[str, Parameter], action: Action):
+        self.name = name
+        self.description = description
+        self.parameters = parameters
+        self.action = action
 
-        return tool
+    def to_str(self) -> str:
```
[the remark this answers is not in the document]
Ch0live: Adjusted in a new branch [here](https://github.com/WaitThatShouldntWork/InferGPT/tree/feature/refactor-director)

PR #57, opened 2024-05-21
Websockets healthcheck

harryBedfordSL:
## Description

This PR adds a basic, single websocket connection to the application. It sets up a heartbeat between the server and client. The server will disconnect the socket if it does not retrieve a `ping` within the heartbeat_timeout 

## Changelog
- Added new websocketProvider on the frontend to allow injection of websocket state into components
- Added new connection status icon in the frontend to show websocket isConnected status
- Added `ping` heartbeat from the frontend -> backend (ping is sent every 10 seconds)
- Moved `api.py` to its own module and introduced `main.py`
- Added `connection_manager.py` in the backend to deal with connecting and disconnecting the socket
- Added `message_handlers.py` to separate out the handling of messages from the connection manager
- Backend docker file updated to use the new entry point (which now has a host argument)
- Basic tests added

## Note

This PR is not changing any of the functionality of inferGPT. It is adding websockets for future use.

## Screenshot

New connection status icon:

image


harryBedfordSL:
> Looks good, general question/observation about ping/pong though...
> 
> I assume these are to both keep the session alive and to make sure data can flow back and forth, however, I think sending a ping every n seconds regardless of what is happening on the connection could become wasteful of network bandwidth.
> 
> While the current approach is ok for now, maybe a future iteration could be that while data is flowing no ping/keepalive is required, but only send a ping when there has been no traffic for the n seconds.

Yeah that sounds like a great iteration to me

philbaxter (dismissed):

backend/tests/api/api_test.py
```diff
@@ -7,7 +7,7 @@
 
 
 def test_health_check_response_healthy(mocker):
-    mock_test_connection = mocker.patch("src.api.test_connection", return_value=True)
+    mock_test_connection = mocker.patch("src.api.api.test_connection", return_value=True)
```
bheyman-scottlogic, line 10: Should we rename the directory or the file so that they are not the same?

frontend/src/app.tsx
```diff
@@ -1,16 +1,21 @@
...
   const { sendMessage, messages, waiting } = useMessages();
 
+  const { isConnected } = useContext(WebsocketContext);
+
   return (
     <div className={styles.container}>
       <Chat messages={messages} waiting={waiting} />
       <Input sendMessage={sendMessage} />
-    </div>
+      <ConnectionStatus isConnected={isConnected} />
```
bheyman-scottlogic, line 18: I would quite like to see the dot inside the chat window area, opposed to the top of the screen. What do you think on this?

frontend/src/session/websocket-provider.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export interface WebsocketProviderProps {
+  children: React.ReactNode;
+}
+
+const WS_URL = process.env.WS_URL || 'ws://localhost:8250/ws';
+const heartbeatInterval = 10000;
+const ping = JSON.stringify({ type: MessageType.PING });
+
+export const WebsocketProvider = ({ children }: WebsocketProviderProps) => {
```
bheyman-scottlogic, line 12: Please could you explain to me what the websocket context and provider are doing?

bheyman-scottlogic (commented):
Great work Harry, just had a couple of questions/suggestions for you

backend/tests/api/api_test.py
harryBedfordSL: yeah that is a good point, maybe it should be called `app.py` as app is what comes from it. I'll change to that, open to other ideas though

frontend/src/app.tsx
harryBedfordSL: Yeah happy to move it

frontend/src/session/websocket-provider.tsx
harryBedfordSL: Discussing offline

philbaxter (approved):

bheyman-scottlogic (approved):

PR #60, opened 2024-05-23
Populate database script

bheyman-scottlogic:
## Description
This PR populates the db with the test data on startup with docker. It also removes the unusable data from the dataset for now. Currently the schema looks like this: 
image
 NB. There is no person node as this was not in the sample data. Happy to add any additional nodes into this if we think that it is necessary, but this is currently based on the data we have.

## Changelog
- Added database population on startup of docker instance
- Added tests
- Added cypher scripts (If you think these should be in a different file location/format, I am happy to update this as well)
- Added test data file (Accounting for about 240,000 lines of the changes)


backend/tests/api_test.py
```diff
@@ -45,3 +46,18 @@ def test_chat_response_failure(mocker):
...
+    mock_annual_transactions_cypher_script = mocker.patch(
+        "src.api.annual_transactions_cypher_script", return_value=(mocker.Mock())
+    )
+    mock_annual_transaction_data = mocker.Mock()
+    mocker.patch("json.load", return_value=mock_annual_transaction_data)
+
+    with client:
+        pass
+
+    mock_populate_db.assert_called_once_with(mock_annual_transactions_cypher_script, mock_annual_transaction_data)
```
harryBedfordSL, line 63: should this assert be inside the `with client`? to avoid needing the pass?

backend/tests/utils/graph_db_utils_test.py
```diff
@@ -32,11 +37,52 @@ def test_database_connectivity_is_healthy(mock_driver):
...
+    mocker.patch("src.utils.graph_db_utils.remove_transactions_without_merchant", remove_transactions_without_merchant)
+    mocker.patch("src.utils.graph_db_utils.remove_connecting_nodes", remove_connecting_nodes)
+
+    populate_db(query, data)
+
+    mock_session.run.assert_any_call("MATCH (n) DETACH DELETE n")
+    mock_session.run.assert_any_call(query, data={"all_data": data})
+    mock_session.run.assert_any_call(remove_credits)
+    mock_session.run.assert_any_call(remove_transactions_without_merchant)
+    mock_session.run.assert_any_call(remove_connecting_nodes)
```
harryBedfordSL, line 70: I don't know what I prefer, so up to you, but you could get the order asserted here too (assuming that `assert_any_call` just looks for any time that the method was called with) with something like in meesage_handlers_test:
```python
first_call = mocked_create_task.call_args_list[0]
assert first_call == call(mock_ws.send_json(pong))
```

backend/src/api/app.py
```diff
@@ -1,21 +1,36 @@
...
+from src.utils.annual_cypher_import import annual_transactions_cypher_script
 
 config_file_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "config.ini"))
 logging.config.fileConfig(fname=config_file_path, disable_existing_loggers=False)
 logger = logging.getLogger(__name__)
 
-app = FastAPI()
+
+@asynccontextmanager
+async def lifespan(app: FastAPI):
```
hsauve-scottlogic, line 22: very interesting to see how lifespan works!

backend/src/utils/annual_cypher_import.py
```diff
@@ -0,0 +1,49 @@
...
+            FOREACH (payment_classification IN t.transaction_classification |
+                MERGE (classification:Classification {name:payment_classification})
+                MERGE (transaction)-[:CLASSIFIED_AS]->(classification)
+            )
+        )
+    )
+)
+"""
+
+remove_credits = """
```
hsauve-scottlogic, line 36: Sorry Ben, what does credits refer to? Is it in the annual transactions data? For some reason I can't view that file

backend/src/utils/graph_db_utils.py
```diff
@@ -43,3 +44,30 @@ def execute_query(llm_query):
         if session:
             session.close()
         driver.close()
+
+
+def populate_db(query, data) -> str:
+    data = {"all_data": data}
+    try:
+        with driver.session() as session:
+            session.run("MATCH (n) DETACH DELETE n")
```
hsauve-scottlogic, line 53: Can the clearing be done after 'yield' in app.py? Just wondering if there are benefits to do it one way or another

hsauve-scottlogic (commented):
Looks really good Ben, it was very interesting to learn about lifespan. Are you thinking of editing the /data README? I added a note about loading data but it won't be relevant once your code is in?

backend/tests/utils/graph_db_utils_test.py
```diff
@@ -1,28 +1,33 @@
...
+    return driver
 
 
 def remove_whitespace_and_newlines(original):
     return " ".join(original.replace(r"\n", " ").replace(r"\r", "").split())
 
 
-def test_database_connectivity_is_healthy(mock_driver):
+def test_database_connectivity_is_healthy(mocker, mock_driver):
+    mocker.patch("src.utils.graph_db_utils.driver", mock_driver)
```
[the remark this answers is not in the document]
bheyman-scottlogic: My understanding is that it needs to be in the individual test file. Have tried with it in the driver fixture as well and this did not work

backend/src/utils/annual_cypher_import.py
bheyman-scottlogic: The credits are anything that is a positive in flow of money ie. winnings from lottery or income. Debits are the negative outflows. These credits are removed to make it work with our schema for now

backend/src/utils/graph_db_utils.py
bheyman-scottlogic: It could be done at the end, yes. Maybe it would be faster for the bootup but I believe it works well at this point for now. I didn't think of doing it that way round, good idea!

backend/src/utils/annual_cypher_import.py
hsauve-scottlogic: Thanks, I understand in the context of the schema now

hsauve-scottlogic (approved):

backend/src/utils/mock_transactions/annual_transactions.json
harryBedfordSL, line 1: Just had a thought, this is quite big. I think I say it was 9Mb when I opened it, which seems large to have in our codebase. What do you think of either:
- Zipping this up (How small would it go?)
- Use Azure Dev Ops to run our CI/CD (maybe that could store this)
- Use Azure Blob storage and retrieve the json file remotely?

philbaxter (approved):

hsauve-scottlogic (approved):

PR #64, opened 2024-06-04
add project setup for pycharm e.g. code style etc

philbaxter:
## Description
Add setup for pycharm IDE including code style etc.

### DIFF
--- a/frontend/.idea/codeStyles/Project.xml
+++ b/frontend/.idea/codeStyles/Project.xml
@@ -1,0 +1,71 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <option name="LINE_SEPARATOR" value="&#10;" />
+    <JSCodeStyleSettings version="0">
+      <option name="FORCE_SEMICOLON_STYLE" value="true" />
+      <option name="USE_DOUBLE_QUOTES" value="false" />
+      <option name="FORCE_QUOTE_STYlE" value="true" />
+    </JSCodeStyleSettings>
+    <JavaCodeStyleSettings>
+      <option name="JD_ADD_BLANK_AFTER_PARM_COMMENTS" value="true" />
+      <option name="JD_ADD_BLANK_AFTER_RETURN" value="true" />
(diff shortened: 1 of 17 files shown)


frontend/.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,71 @@
...
+      <option name="INDENT_CASE_FROM_SWITCH" value="false" />
+      <option name="ALIGN_MULTILINE_PARAMETERS" value="false" />
+      <option name="ALIGN_MULTILINE_FOR" value="false" />
+      <indentOptions>
+        <option name="INDENT_SIZE" value="2" />
+        <option name="CONTINUATION_INDENT_SIZE" value="2" />
+      </indentOptions>
+    </codeStyleSettings>
+  </code_scheme>
+</component>
```
harryBedfordSL, line 71: are all these captured in the `eslintrc.json` ? Feels like the IDE shouldn't be doing the linting, but pointing to `eslint`

philbaxter: These are the built-in pycharm checks, I am guesing this is whats done by the editor rather than checking at build time

harryBedfordSL: But could this contradict something that we add to the linter? (I'm wondering whether IDE specific code should be in the repo at all, we are git ignoring the `.vscode` directory at the moment)

philbaxter: I think if you get the IDE to check as you are editing then you will have less chance of github actions throwing errors at you, resulting in fewer commits that just fix style issues! Also, if we commit IDE settings (pycharm or vscaode) to github, then everybody is going to be singing off the same hymnsheet as the code is written.

harryBedfordSL: Yeah that's fair, I can get on board with that. In that case, I think we should also push the `.vscode/settings.json` file to reflect the same as your changes here (in a separate PR / ticket 🙂)

harryBedfordSL (approved):

bheyman-scottlogic (approved):

PR #65, opened 2024-06-05
Add semantic layer to graph database schema

hsauve-scottlogic:
## Description

This PR incorporates [Michael's colab code](https://colab.research.google.com/drive/165hS_kknWiFQNKBQbqcHiYu-ayFs2JEl?usp=sharing#scrollTo=d83nhXipRZB1) into our DataStoreAgent. The issue we had was that the graph schema was not fleshed out enough to give the context the LLM needed to understand the structure of the graph.

By enriching the schema with metadata on each element of the graph, the DataStoreAgent has more context to generate Cypher queries and is less prone to hallucinations

## Changelog
- Added APOC plugin in Compose file
- Added 4 Cypher scripts to fetch nodes, relationships, node properties and relationships
- Added 4 prompts for each of the above to get the LLM to construct an understanding of each element
- Moved the returned value of graph_shema into a jinja template to feed it statically to the DataStoreAgent
- Moved the semantic_layer_builder to utils as it is not currently needed


philbaxter (approved):
Great work, some duplication in the j2 files that we may be able to remove, but we should do that as part of refactoring once we have everything assembled together.

backend/src/prompts/templates/neo4j-graph-why.j2
```diff
@@ -0,0 +1,4 @@
+The graph represents a hierarchy starting from a Parent Account,
+which is at the top level of the graph and goes down to Transaction.
+All Transactions belong to a group or Classification and each of them
```
bheyman-scottlogic, line 3: This is a little confusing to me. Could maybe be made clearer by explaining the relationships and a short explanation of each thing.

backend/src/prompts/templates/neo4j-node-property.j2
```diff
@@ -0,0 +1,15 @@
+Persona: You are a graph modelling expert who understands the nuances of the graph model currently loaded into Neo4j.
+
+Role: You must update documentation with the details around the specific property on each node in the graph.
+Focus on the property as a priority
+
+Additional Info:
+- The length of all vector embeddings is 256 dimensions
```
bheyman-scottlogic, line 7: Do we have vector embeddings in this?

backend/src/prompts/templates/neo4j-property-intent-prompt.j2
```diff
@@ -0,0 +1,12 @@
+Persona: You are a graph modelling expert who understands the nuances of the graph model currently loaded into Neo4j.
+
+Role: You must update documentation with the details around the specific property on each relationship in the graph.
+Focus on the property as a priority
```
bheyman-scottlogic, line 4: I don't see how this is different to the `neo4j-relationship-understanding.j2` file. Why do we need all of these separate as well? We could just have the word as an arguement that gets changed when we build the prompt

backend/src/utils/semantic_layer_builder.py
```diff
@@ -0,0 +1,78 @@
...
+def get_semantic_layer():
+    finalised_graph_structure = {'nodes':{}, 'properties':{}}
+
+    neo4j_graph_why_prompt = engine.load_prompt("neo4j-graph-why")
+
+    relationship_query = engine.load_prompt("relationships-query")
+
+    node_query = engine.load_prompt("nodes-query")
+
+    relationship_property_query = engine.load_prompt("relationship-property-cypher-query")
```
bheyman-scottlogic, line 23: For comment above: add in an arguement here and can then just load the same prompt eachtime with a different arguement to differentiate

backend/src/utils/semantic_layer_builder.py
```diff
@@ -0,0 +1,78 @@
...
+    neo4j_node_property_prompt = engine.load_prompt("neo4j-node-property",
+                                                    neo4j_graph_why_prompt=neo4j_graph_why_prompt)
+
+
+    # Fetch and enrich relationships
+    relationship_result = execute_query(relationship_query)
+    relationships_neo4j = relationship_result[0]
+    enriched_relationships = call_model(neo4j_relationships_understanding_prompt, str(relationships_neo4j))
+    enriched_relationships = json.dumps(enriched_relationships)
+    enriched_relationships = json.loads(enriched_relationships)
```
bheyman-scottlogic, line 45: Do we need to do JSON dumps and JSON loads here?

backend/src/utils/semantic_layer_builder.py
```diff
@@ -0,0 +1,78 @@
...
+    finalised_graph_structure['relationships'] = enriched_relationships if enriched_relationships else {}
+    logger.info("Finalised graph structure with enriched relationships: ")
+    logger.info(finalised_graph_structure)
+
+    # Fetch and enrich nodes
+    nodes_neo4j_result = execute_query(node_query)
+    nodes_neo4j = nodes_neo4j_result[0]
+    enriched_nodes = call_model(neo4j_nodes_understanding_prompt, str(nodes_neo4j))
+    enriched_nodes = ast.literal_eval(enriched_nodes)
```
bheyman-scottlogic, line 54: What does this do?

bheyman-scottlogic (commented):
Great work and very cool to see it working! Just some comments for clarification and some small changes. Nice one Hélène

backend/src/agents/datastore_agent.py
```diff
@@ -8,28 +8,28 @@
...
             type="string",
             description="The utterance from the user that the query will be based on",
         )
     }
 )
 def generate_query(user_prompt) -> str:
+        generate_cypher_query_prompt = engine.load_prompt("generate-cypher-query",
+                                                  graph_schema=graph_schema,
+                                                  current_date=datetime.now())
+        print("generate cypher query prompt: " + generate_cypher_query_prompt)
```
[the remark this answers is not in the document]
harryBedfordSL: will remove, don't think it is needed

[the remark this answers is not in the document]
harryBedfordSL: actually, am going to make a follow up ticket to do the refactoring. Will capture your comments there

harryBedfordSL (dismissed):
Looks great to me Hélène! I've made a follow up ticket to address what we've discussed. Ben , I've captured some of your comments too in the new ticket

backend/src/utils/semantic_layer_builder.py
harryBedfordSL: my understanding is that it is converting a string safely into a python dictionary. But Hélène please correct me if I'm wrong

bheyman-scottlogic (approved):
Im happy if these comments are being addressed in the new ticket, great stuff!

philbaxter (approved):

backend/src/utils/semantic_layer_builder.py
hsauve-scottlogic: Yes that's my understanding. enriched_nodes is already a string representation of a Python dictionary. By using ast (Abstract Syntax Trees) we can safely evaluate the string as a Python object (without it, I was getting json errors)

PR #66, opened 2024-06-05
Wire in Intent Agent 

bheyman-scottlogic:
## Description
Bring in Intent Agent to flow in replacement of the `create_tasks` agent

## Changelog
- Wire in intent agent
- Update director
- Update supervisor
- Update datastore agent
- Add `answer_agent`


backend/src/agents/answer_agent.py
```diff
@@ -0,0 +1,22 @@
+from datetime import datetime
+from typing import Optional
+from src.utils import call_model
+from src.prompts import PromptEngine
+from src.agents import Agent, agent
+
+engine = PromptEngine()
+intent_format = engine.load_prompt("intent-format")
```
harryBedfordSL, line 8: Don't think this prompt is needed in this file?

backend/src/agents/answer_agent.py
```diff
@@ -0,0 +1,22 @@
...
+
+datetime = datetime.now()
```
harryBedfordSL, line 10: This may be in the slightly wrong position. I wonder whether this would be computed once when this agent is imported to the director. Would we want it to be the exact datetime when the answer agent is invoked, and if so then I feel like this should be inside the invoke? (I may be barking up the wrong tree hear!)

backend/src/director.py
```diff
@@ -14,17 +14,13 @@
...
-    if determine_intention(question) == "TRUE":
-        task_dict = create_tasks(question, agents_details)
-        final_answer = solve_all_tasks(task_dict)
-        return final_answer
-
-    logging.info("Passing utterance straight to call_model function")
-    return call_model(director_prompt, user_prompt=question)
+    intent_json = json.loads(intent)
+    logging.info(f"Intent determined: {intent}")
+    final_scratchpad = solve_all_tasks(intent_json)
```
harryBedfordSL, line 20: I wonder whether `solve_all_tasks` should be renamed to `solve_all` so that it is agnostic of tasks vs questions

backend/src/supervisors/supervisor.py
```diff
@@ -25,30 +25,29 @@ def solve_task(task, scratchpad, attempt=0) -> Tuple[str | None, str]:
     if agent is None:
         return (None, no_agent_response)
 
-    answer = agent.invoke(task["summary"])
-    if is_valid_answer(answer, task):
-        return (agent.name, answer)
+    answer = agent.invoke(task)
+    # if is_valid_answer(answer, task):
```
harryBedfordSL, line 29: hmm, I wonder if the validator agent needs to be more light touch for now, and then something we can experiment on in the future e.g. do we get an answer, if so great lets say success ?

backend/src/agents/datastore_agent.py
```diff
@@ -16,29 +16,57 @@
...
-        logger.info(json_query)
-        if json_query["query"] == "None":
-            return "No database query"
-        db_response = execute_query(json_query['query'])
-        logger.info(db_response)
-        return str(db_response)
+def generate_query(question_intent, operation, question_params, aggregation, sort_order, timeframe) -> str:
+    details_to_create_cypher_query = (
+        f"{question_intent} {operation} {question_params} {aggregation} {sort_order }{timeframe}"
+    )
```
harryBedfordSL, line 60: Should this be in a prompt template with some description around each variable?

backend/src/prompts/templates/best-tool.j2
```diff
@@ -16,6 +16,8 @@ Add if appropriate, but do not hallucinate arguments for these parameters
 
 {{ tools }}
 
+From the task you should be able to extract the parameters. If it is data driven, it should be turned into a cypher query
```
philbaxter, line 19: Should we specify a specific agent for this, or can it work it out for itself?

bheyman-scottlogic: This is selecting the tool within the agent, so the agent has already been selected. Based on this, it can work it out itself. I don't think its 100% there but it is working for now. Before having this it wouldn't be able to select the right tool, even though there was one tool. I reckon some tweaking could improve this after adding in the semantic work as well

philbaxter (approved):

harryBedfordSL (approved):

PR #72, opened 2024-06-20
Feature/refactor backend tests env dependency

harryBedfordSL:
## Description

This PR removes the dependency that our tests have on the .env file. This arose from certain imports causing files to run that were making use of env variables. My understanding is that this is a compile time issue. So, the change has been to make sure that wherever the env variables are used, it is at run time, and not compile time. For example, using getters for the agents instead of instantiating them in the __init__ file.

## Changelog
- Decouple unit test dependency to .env file
- Delete creation of .env file in ci

### DIFF
--- a/.github/workflows/test-backend.yml
+++ b/.github/workflows/test-backend.yml
@@ -24,26 +24,6 @@ jobs:
           pip install -r requirements.txt
           pip install pytest-md pytest-emoji
 
-        # some tests require an .env file to run.
-        # file should contain all the required properties.  
-      - name: Create environment file
-        run: |
-          echo "Create environment file"
-          cat <<EOF > .env
-          MODEL=my-model
-          MISTRAL_KEY=123456
-          NEO4J_URI=bolt://localhost:1234
-          NEO4J_USERNAME=neo4j
-          NEO4J_PASSWORD=my-password
-          ANSWER_AGENT_LLM=mistral
-          INTENT_AGENT_LLM=mistral
-          VALIDATOR_AGENT_LLM=mistral
-          DATASTORE_AGENT_LLM=mistral
-          MATHS_AGENT_LLM=mistral
-          ROUTER_LLM=mistral
-          EOF
-          cat .env
-
       - name: Run tests
         uses: pavelzw/pytest-action@v2
         with:
--- a/backend/src/agents/__init__.py
+++ b/backend/src/agents/__init__.py
@@ -27,12 +27,12 @@ def agent_details(agent) -> dict:
     return {"name": agent.name, "description": agent.description}
 
 
-def get_question_agents() -> List[Agent]:
+def get_available_agents() -> List[Agent]:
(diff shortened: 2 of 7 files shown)


backend/src/agents/__init__.py
```diff
@@ -10,26 +11,40 @@
...
+    return AnswerAgent(config.answer_agent_llm)
+
+
+def agent_details(agent) -> dict:
     return {"name": agent.name, "description": agent.description}
 
 
-agents = [DatastoreAgent(config.datastore_agent_llm), MathsAgent(config.maths_agent_llm)]
-agents_details = [get_agent_details(agent) for agent in agents]
+def get_question_agents() -> List[Agent]:
```
hsauve-scottlogic, line 30: I wonder if the name question_agents is too vague, could it be something like available_agents?

backend/tests/router_test.py
```diff
@@ -24,8 +26,9 @@ def test_get_agent_for_task_no_agent_found(mocker):
 
 def test_get_agent_for_task_agent_found(mocker):
     plan = {"agent_name": mock_agent_name}
-    mocker.patch("src.router.agents", mock_agents)
-    mocker.patch("src.router.llm", mock_model)
+    mocker.patch("src.router.get_llm", return_value=mock_model)
```
hsauve-scottlogic, line 29: Just so I understand, we're not actually using the get_llm function which needs the env file, instead we're instantiating a mock value for llm with 'mockllm' on line 8?

hsauve-scottlogic (commented):
Looks great Harry! I've asked a couple of questions, thanks

backend/src/agents/__init__.py
harryBedfordSL: yeah I like available_agents, will change to that

backend/tests/router_test.py
harryBedfordSL: Actually, this is a little confusing, I'll change. Yes, the test is mocking out `get_llm` so that whenever it is called during the test, the mock_model on line 8 will always be returned. The confuding part is that line 8 is actually using `get_llm` to get the mock model. It would make more sense if line 8 was:
```python
mock_model = MockLLM()
```

.github/workflows/test-backend.yml
gaganahluwalia, line 1: @harryBedfordSL  - I am unable to test it locally. Even after renaming the .env file locally, all the tests are running in my other branch and according to you they should fail without all the changes in this PR, so not sure how to test this one.

backend/tests/router_test.py
hsauve-scottlogic: Ah yes okay, thanks that makes sense

gaganahluwalia (changes requested):
Happy with the changes.

gaganahluwalia (approved):

hsauve-scottlogic (approved):

PR #75, opened 2024-07-01
Remove model env variable and add model in for each agent

hsauve-scottlogic:
## Description

This PR removes the need for the MODEL env variable. Instead of having the LLM model defined in Mistral, the model is defined within each agent. This way, we can use different model sizes for specific tasks. 

## Changelog

- Changed the signature of the chat method to have a model argument
- Removed the MODEL env variable
- Added MODEL env variables for each agent
- Tweaked tests to add in model
- Edited env.example with new variables


.env.example
```diff
@@ -38,3 +37,10 @@ VALIDATOR_AGENT_LLM="mistral"
 DATASTORE_AGENT_LLM="mistral"
 MATHS_AGENT_LLM="mistral"
 ROUTER_LLM="mistral"
+
```
gaganahluwalia, line 40: Will make sense to put a comment # llm model here.

backend/src/agents/agent.py
```diff
@@ -37,7 +38,7 @@ def __get_action(self, utterance: str) -> Action_and_args:
 
         logger.debug(f"List of tools: {tool_descriptions}")
 
-        response = json.loads(self.llm.chat(format_prompt, tools_available))
+        response = json.loads(self.llm.chat("mistral-large-latest", format_prompt, tools_available))
```
gaganahluwalia, line 41: mistral model is hard coded here. It should come from the config.

compose.yml
```diff
@@ -60,6 +60,11 @@ services:
       DATASTORE_AGENT_LLM: ${DATASTORE_AGENT_LLM}
       MATHS_AGENT_LLM: ${MATHS_AGENT_LLM}
       ROUTER_LLM: ${ROUTER_LLM}
+      ANSWER_AGENT_MODEL: ${ANSWER_AGENT_MODEL}
+      INTENT_AGENT_MODEL: ${INTENT_AGENT_MODEL}
+      VALIDATOR_AGENT_MODEL: ${VALIDATOR_AGENT_MODEL}
+      DATASTORE_AGENT_MODEL: ${DATASTORE_AGENT_MODEL}
+      ROUTER_MODEL: ${ROUTER_MODEL}
```
gaganahluwalia, line 67: MATHS_AGENT_MODEL is missing here.

backend/src/agents/agent.py
hsauve-scottlogic: Yes I hardcoded it here because the agent class doesn't have a default model, but I appreciate it's not ideal. I'll create an agent class model env variable, unless you have another idea

gaganahluwalia: I think we need to create an env variable otherwise once we start using openai or any other LLM in combination with this agent.py it may fail as I am not sure how it's going to work in combination with any other LLM.

gaganahluwalia (dismissed):

philbaxter (dismissed):

gaganahluwalia (approved):

PR #76, opened 2024-07-02
OpenAI changes for the ticket 112

gaganahluwalia:
## Description
This PR involves integrating OpenAI into the InferGPT project. The changes include modifications to support OpenAI's API for language model interactions, allowing the system to utilize OpenAI's capabilities for various tasks.

## Changelog
- Added: OpenAI client and integration into the existing LLM framework.
- Unit tests for OpenAI integration.
- Configured: Environment variables for OpenAI API key and model selection.
- Updated requirements.txt with the OpenAI package 1.35.3
- Updated config.py to include openAI key and model
- Updated compose.yml to include openAI key and model


philbaxter (approved):
Good job, not necessary for now, but we may want to think about putting temperature and max_tokens in the .env file at some point.

backend/src/llm/openai_client.py
```diff
@@ -0,0 +1,27 @@
...
+    def __init__(self, api_key: str):
+        self.api_key = config.openai_key
+        openai.api_key = self.api_key
+
+    def chat(self, model, messages, temperature=0, max_tokens=150):
+        try:
+            response = openai.ChatCompletion.create(
+                model=model,
+                messages=messages,
+                temperature=0,
```
philbaxter, line 19: Hmm, am I seeing this correctly, you are passing in to the function temp and max tokens, but then hardcoding them here?

philbaxter (changes requested):

backend/src/llm/openai_client.py
gaganahluwalia: @philbaxter  - I have removed the hard coded values.

philbaxter: Hmm, still not sure its quite right, what are the two values you now pass into the chat method used for? I was expecting the two parameters passed into chat() to then be passed on to the create method, but it seems you have now removed them.

gaganahluwalia: @philbaxter  - The current 2 values passed are model & messages. Earlier I was passing Temp and Tokens but those have default values in the chat method.
i.e.
def chat(self, model, messages, temperature=0, max_tokens=150):

backend/requirements.txt
```diff
@@ -16,3 +16,4 @@ cffi==1.16.0
 cryptography==42.0.7
 isodate==0.6.1
 pycparser==2.22
+openai==1.35.3
```
hsauve-scottlogic, line 19: just needs an extra line here

hsauve-scottlogic (commented):
Looks great, nice work Gagan! Also good conversation about future potential refactoring. Works on my machine and tests pass

backend/requirements.txt
gaganahluwalia: Added the extra line

backend/src/llm/openai_client.py
gaganahluwalia: @philbaxter  - Please have a look now and see if it looks good to you.

hsauve-scottlogic (approved):

philbaxter (approved):

PR #77, opened 2024-07-08
Rewrite Connection Manager to Handle Multiple Connections

swood-scottlogic:
## Description
This PR just handles the changes made to the existing web socket code. The previous implementation made it very difficult to send data on the web socket unless it was in response to a message received. The new implementation starts to handle the idea of having multiple connections active at one time, but will require more work to be done on the implementation of user sessions.

## Changelog
- Reworks connection manager to handle multiple web sockets
- Adds new broadcast method to send Messages from the backend on the web socket
- Removes on_message method and recreates functionality in app.py
    - Functionality changed from using recursion to a while loop as long running connections could cause a recursion depth exceeded exception
- Rewrites unit tests to better cover new structure

### DIFF
--- a/backend/tests/api/connection_manager_test.py
+++ b/backend/tests/api/connection_manager_test.py
@@ -1,5 +1,9 @@
+import json
+from unittest.mock import patch
 import pytest
 from fastapi.websockets import WebSocketState
+from src.api.types import Message, MessageTypes
+from src.api.message_handlers import handlers
 from src.api.connection_manager import ConnectionManager
 
 error_message = "Test Error"
@@ -14,8 +18,8 @@ def connection_manager(mocker):
     manager = ConnectionManager()
     mock_ws = mocker.AsyncMock()
     mock_ws.accept = mocker.AsyncMock()
+    mock_ws.send_json = mocker.AsyncMock()
     mock_on_message = mocker.AsyncMock
-    mocker.patch.object(manager, "_ConnectionManager__on_message", new_callable=mock_on_message)
 
     return manager, mock_ws, mock_on_message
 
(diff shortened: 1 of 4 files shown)


gaganahluwalia (dismissed):
Looks all good to me.

backend/tests/api/connection_manager_test.py
```diff
@@ -1,84 +1,159 @@
...
+    mock_ws.accept = mocker.AsyncMock()
+    mock_ws.send_json = mocker.AsyncMock()
+    mock_on_message = mocker.AsyncMock
+
+    return manager, mock_ws, mock_on_message
+
+
+@pytest.mark.asyncio
+async def test_connect_new_websocket(connection_manager):
+    manager, mock_ws, _ = connection_manager
```
hsauve-scottlogic, line 29: what does the underscore represent here please?

hsauve-scottlogic (approved):
Looks good to me, and all tests pass on my machine, I just left one comment for clarification, thanks

backend/tests/api/connection_manager_test.py
swood-scottlogic: The `connection_manager` function defined on line 17 returns three things: `manager, mock_ws, mock_on_message`. In this test I don't need the `mock_on_message` value that has been returned and one convention to signify I won't be using this is to assign it to the variable `_`. I've had a look and can't find an official source about this convention but it is referenced in various places. The additional benefit is pylance (the language server vscode uses for warnings and errors in python)  won't report the `_` variable as being unused.

gaganahluwalia (approved):

hsauve-scottlogic (approved):

PR #80, opened 2024-07-22
Feature/setup pyright ci

swood-scottlogic:
## Description
Adds a GitHub action that carries out static type checking of our backend code and also fixes existing issues throughout the codebase.

## Changelog
- Adds a pyright config file to standardise across developers type checking settings that can also be used by CI.
-Adds a github action to run [pyright](https://github.com/microsoft/pyright) static type checker on our backend code using the pyright config file.
-Fixes following flagged errors
  - Argument Type errors
  - Possible Unbound Variable errors
  - Incompatible Method Override errors
- Ignores Attribute Access Issues as this error was only reported in code not used at run time

### DIFF
--- a/.github/workflows/type-check-backend.yml
+++ b/.github/workflows/type-check-backend.yml
@@ -1,0 +1,28 @@
+name: Type Check Backend
+on:
+  pull_request:
+
+jobs:
+  checking:
+    name: Type Checking Backend
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Install dependencies
+        run: |
+          cd ./backend
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+
+      - name: Type Check Backend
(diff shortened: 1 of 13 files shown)


philbaxter (approved):
Good job

.github/workflows/type-check-backend.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Install dependencies
```
hsauve-scottlogic, line 18: Because we install dependencies in test-backend as well, I wonder if there's a way to cache that bit to reduce time and network utilisation, but that could be one for future improvements

hsauve-scottlogic (approved):
Looks great Seth! Just made one comment but as a thought rather than a change. Nice bit of tidying up

.github/workflows/type-check-backend.yml
swood-scottlogic: Yeah I think it's a trade off situation. We definitely could cache it but that would probably entail us linking the two actions together somehow or even combining them into one. We could definitely do that, but it's possible not being able to run the two actions in parallel could end up being slower.

PR #81, opened 2024-07-24
Feature/move llms to async clients

swood-scottlogic:
## Description
Moves as much of the I/O bound code as possible over to using async implementations.
This solves the problem of only a singular request being able to be worked on at any one time and also allows websocket messages to correctly send messages from the api during computation of the request

## Changelog
- Move Existing Code to Async Implementations
   - In both mistral.py and openai.py the current clients are replaced
with async versions. This causes a knock on effect of a lot of our
code being altered to be async to make use of the llms.
- Call async log publisher method directly and remove synchronous method
  - Now that everywhere we were calling the log publisher has been moved over to async implementation we can call the async log publisher methods directly


philbaxter (commented):
*I think* it looks ok, a lot there to check. Do we need to put _async on the log method? as all we will be allowing is an async call, it seems a bit superfluous to me

swood-scottlogic:
> _I think_ it looks ok, a lot there to check. Do we need to put _async on the log method? as all we will be allowing is an async call, it seems a bit superfluous to me

Yeah definitely agree it's a lot to check! Unfortunately there wasn't really a way to split it so that it still worked once I started moving things over to async.

Good spot on the _async suffix, that was there when we had the async and sync versions of the method. I'll rename the function now to not have the _async suffix :)

philbaxter (approved):

hsauve-scottlogic (commented):
Looks great Seth! I just had one thought/question

backend/src/agents/intent_agent.py
```diff
@@ -11,7 +11,7 @@
     tools=[],
 )
 class IntentAgent(Agent):
-    def invoke(self, utterance: str) -> str:
+    async def invoke(self, utterance: str) -> str:
         user_prompt = engine.load_prompt("intent", question=utterance)
 
-        return self.llm.chat(self.model, intent_format, user_prompt=user_prompt, return_json=True)
+        return await self.llm.chat(self.model, intent_format, user_prompt=user_prompt)
```
gaganahluwalia, line 17: So we don't want the return_json = True anymore?

backend/src/agents/web_agent.py
```diff
@@ -15,48 +15,53 @@
...
 
-def is_valid_answer(answer, task) -> bool:
-    is_valid = (get_validator_agent().invoke(f"Task: {task}  Answer: {answer}")).lower() == "true"
+
+async def is_valid_answer(answer, task) -> bool:
+    is_valid = (await get_validator_agent().invoke(f"Task: {task}  Answer: {answer}")).lower() == "true"
     return is_valid
 
+
 def perform_search(search_query: str, num_results: int) -> Dict[str, Any]:
```
gaganahluwalia, line 65: Any reason why this function i.e. perform_search is nor async?

backend/src/agents/intent_agent.py
swood-scottlogic: good catch, I had to resolve a lot of merge conflicts and must have made a mistake here. Will add return_json=True back in

backend/src/agents/web_agent.py
swood-scottlogic: In most cases I've only made functions async if they use something that needs to be awaited. (There are some exceptions but those are to do with inheriting a method that is expected to be async e.g the `mock_llm` `chat` method). The search_urls method we call here makes use of the `googlesearch` library which as far as I can see does not have support for async.

backend/src/agents/intent_agent.py
swood-scottlogic: Have pushed this change now :)

gaganahluwalia (approved):
Happy with the changes.

hsauve-scottlogic (approved):

PR #79, opened 2024-07-18
Feature/create web agent

gaganahluwalia:
## Description
This PR involves integrating a new **'WebAgent'** into the **InferGPT** project. The changes include adding the **WebAgent** for handling web search queries and summarizing information from the web. Several files have been modified to support this new functionality.

## Changelog
**WebAgent**

-   Added web_agent.py for the new WebAgent class and tools.
-   Created prompt templates for the summarizer in summariser.j2.
-   Created utility functions in web_utils.py for searching and scraping web content.
**Tools**

- Added the web_general_search tool for performing web searches and summarizing the results.

**Updated**

- Added WEB_AGENT_LLM and WEB_AGENT_MODEL to .env.example and compose.yml.
- Added beautifulsoup4==4.12.3, requests==2.32.3, and googlesearch-python==1.2.4 to requirements.txt.
- Updated director.py to create an instance of WebAgent and handle intents related to web searches.
- Updated config.py to include WEB_AGENT_LLM and WEB_AGENT_MODEL.
- Modified __init__.py to include WebAgent in the list of agents.
- Added examples of web search-related intents in intent.j2.

PR #82, opened 2024-07-25
Add optional parameter to enable json_object response format

swood-scottlogic:
## Description
After an update to the Mistral model, our current prompts were leading to Mistral responses where the generated json was wrapped as if to use it in a .md file. This PR adds an optional parameter to our llm calls that enables or disables json_object response mode for both OpenAI and Mistral which when enabled forces the llm to respond with a valid json response.


pyrightconfig.json
```diff
@@ -10,5 +10,5 @@
     ],
     "typeCheckingMode": "standard",
     "reportMissingImports": true,
-    "reportMissingTypeStubs": true,
+    "reportMissingTypeStubs": false,
```
swood-scottlogic, line 13: The web agent PR (#79) added a package which does not have type information or a stub file with types. I've set this to false for now but will circle back in a follow-up PR to add the stub files for the package and then re-enable this.

backend/src/utils/web_utils.py
```diff
@@ -12,64 +12,38 @@
...
-        )
-        response = llm.chat(model, summariser_prompt, "")
-        return json.dumps({
-            "status": "success",
-            "response": response,
-            "error": None
-        })
+        summariser_prompt = engine.load_prompt("summariser", question=search_query, content=contents)
+        response = llm.chat(model, summariser_prompt, "", return_json=True)
+        return json.dumps({"status": "success", "response": response, "error": None})
```
swood-scottlogic, line 46: These were automatically flattened by the ruff formatter extension. If we'd rather keep these on separate lines it might be worth investigating the ruff config?

philbaxter: I would prefer this split over several lines for readability, especially when the json is a large structure. Could you raise a ticket so we dont lose this investigation/change?

philbaxter (approved):
Just one comment, but looks good.

backend/src/utils/web_utils.py
swood-scottlogic: Yeah definitely!
I've made this [ticket](https://scottlogic.atlassian.net/browse/IG-137) to track this

gaganahluwalia (approved):
Looks all good.

PR #84, opened 2024-08-12
Changing env to yaml

gaganahluwalia:
## Description
This PR involves changing .env to env.yml.  Several files have been modified to support this functionality

## Changelog
- gitignore - Added env.yml in it.
- README.md - Made the required changes to use env.yml instead of .env
- Dockerfile - Had to make the changes to use the backend folder.
- conftest.py - Included a new testing conf file, which sets the variable when pytest is run so that call to env.yml file is ignored.
- requirements.txt - Included pyyml package to read yml files.
- config.py - Made the changes to use variables from the env.yml instead of .env
- compose.yml - Changed the context folder from backend to .
- env.yml.example - This new file is included to provide an example of a sample env.yml file.


env.yml.example
```diff
@@ -0,0 +1,45 @@
...
+backend_url: http://localhost:8250
+ws_url: ws://localhost:8250/ws
+
+# Azure
+azure:
+  storage_connection_string: "my-connection-string"
+  storage_container_name: my-container-name
+  initial_data_filename: test-data.json
+
+# llm config
```
philbaxter, line 26: Might have been nicer to merge the Agents, LLMs and Models e.g.
```
agents:
  - agent: "answer"
    llm: "mistral"
    model: "mistral-large-latest"
  - agent: intent"
    llm: "mistral"
    model: "mistral-large-latest"
```
etc. etc.

This would keep everything for an agent together so you wouldn't need to scroll up and down and will help to not miss something out when adding new agents.

env.yml.example
```diff
@@ -0,0 +1,45 @@
+# neo4j authentication
+neo4j:
+  username: neo4j
+  password: change-me!
+  uri: bolt://localhost:7687
+  http_port: 7474
+  bolt_port: 7687
+
+# backend LLM properties
```
philbaxter, line 9: Maybe reduce the keeys to something like...

```
llms
  - llm: "mistral"
    key: "sgfzdfgzdf"
  - llm: "openai"
    key: "oijojpojih"
```

or something similar

philbaxter (changes requested):
Nice work, just some questions/suggestions about the yml layout that *I think* will make it easier to navigate and read.

not precious about my suggestions though, feel free to ignore and I will approve the PR.

backend/Dockerfile
```diff
@@ -5,13 +5,16 @@ FROM python:3.12
...
 
 # Install the dependencies from the requirements file
 RUN pip install --no-cache-dir --upgrade -r /backend/requirements.txt
 
 # Copy the source code into the working directory
-COPY ./src/. ./src
+COPY backend/src/. ./src
+
+# Copy the env.yaml file from the root directory to the backend directory
+COPY env.yml ./env.yml
```
hsauve-scottlogic, line 17: why does the env file have to move to the backend directory?

hsauve-scottlogic (commented):
Looks great Gagan, I just have one question

backend/Dockerfile
gaganahluwalia: It's because the docker is set up for backend to copy all the files before initialising the project, earlier as we were using .env which gets auto loaded as an environment file and made available using dotenv library. There is no similar library to load yml files as such, so I have to manually move it to the backend folder otherwise the config file is unable to find it.

hsauve-scottlogic (approved):

env.yml.example
gaganahluwalia: @philbaxter - I have made the changes according to your suggestion.

philbaxter (approved):
Nice, looks really good now and the yml is more readable

hsauve-scottlogic (approved):

PR #91, opened 2024-09-09
Featuer/change web agent tests

gaganahluwalia:
## Description
This PR is to upgrade web_agent_test.py as the earlier tests were passing even when there was a change

## Changelog
- 
web_agent_test.py is only file changed


hsauve-scottlogic (approved):
Nice work Gagan

PR #93, opened 2024-09-13
Feature/human in the loop confirmation

swood-scottlogic:
## Description

Adds the ability for agents to get the user to confirm they're happy to continue during their computation. This is currently implemented in the chart generator agent after the code has been generated but before it's executed.

## Changelog
- Moves tests related to files in the websocket folder to mirror folder structure
- Adds the user_confirmer and confirmations_manager classes to the backend which power the confirmation process
- Use the user_confirmer in chart_generator_agent to get confirmation before generating the graph
- Fixes a bug in the backend websocket manager that meant json payloads were being sent as strings not json
- Creates a modal on the frontend to allow the user to send a response to the confirmation request

### DIFF
--- a/frontend/src/components/chat.tsx
+++ b/frontend/src/components/chat.tsx
@@ -3,22 +3,36 @@ import { Message, MessageComponent } from './message';
 import styles from './chat.module.css';
 import { Waiting } from './waiting';
 import { ConnectionStatus } from './connection-status';
-import { WebsocketContext, MessageType } from '../session/websocket-context';
-
+import { WebsocketContext, MessageType, Message as wsMessage } from '../session/websocket-context';
+import { Confirmation, ConfirmModal } from './confirm-modal';
 export interface ChatProps {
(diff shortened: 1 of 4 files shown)


backend/src/websockets/connection_manager.py
```diff
@@ -46,11 +46,12 @@ async def handle_message(self, ws: WebSocket, message: Message):
     async def broadcast(self, message: Message):
         for ws in self.websockets:
             if ws.application_state == WebSocketState.CONNECTED:
-                await ws.send_json(json.dumps({"type": message.type.value, "data": message.data}))
+                await ws.send_json({"type": message.type.value, "data": message.data})
```
gaganahluwalia, line 49: Do you think we should change from json.dumps to normal in other agents as well?

SvetA1395 (approved):

gaganahluwalia (approved):
Looks good.. Well done!

PR #95, opened 2024-09-20
Update mistralai package: v0.1.8 => v1.1.0

swood-scottlogic:
## Description
Update mistralai package to latest version. This update did have breaking changes that caused changes to `mistral.py`

## Changelog
- Update the package in `requirements.txt`
- Update code in `mistral.py` to work after breaking changes
- Add additional tests for new handled error cases after the update

### DIFF
--- a/backend/src/llm/mistral.py
+++ b/backend/src/llm/mistral.py
@@ -30,5 +30,5 @@ class Mistral(LLM):
             logger.error("Call to Mistral API failed: message content is None or Unset")
             return "An error occurred while processing the request."
 
-        logger.debug('{0} response : "{1}"'.format(model, response.choices[0].message.content))
+        logger.debug('{0} response : "{1}"'.format(model, content))
         return content


backend/src/llm/mistral.py
```diff
@@ -9,21 +8,36 @@
...
             messages=[
-                ChatMessage(role="system", content=system_prompt),
-                ChatMessage(role="user", content=user_prompt),
+                SystemMessage(content=system_prompt),
+                UserMessage(content=user_prompt),
             ],
             temperature=0,
             response_format={"type": "json_object"} if return_json else None,
         )
+        if response is None:
```
gaganahluwalia, line 24: Do we need to have separate if checks for Response None and Response.choices?
Won't be easier to just combine these 2,

if not response or not response.choices:
            logger.error("Call to Mistral API failed: No valid response or choices received")
            return "An error occurred while processing the request."

and then we can just proceed with getting the choices

        content = response.choices[0].message.content if response.choices else None
        
        if not content:
            logger.error("Call to Mistral API failed: message content is None or unset")
            return "An error occurred while processing the request."
        // log the content and then just process with returning that content
        logger.debug(f"{model} response: \"{content}\"")
        return content

backend/tests/llm/mistral_test.py
```diff
@@ -58,9 +45,90 @@ async def test_chat_calls_client_chat(mocker):
...
+
+@pytest.mark.asyncio
+async def test_chat_response_none_logs_error(mocker, caplog):
+    mistral.client = mocker.AsyncMock(return_value=mock_client)
+    mistral.client.chat.complete_async.return_value = None
+
+    response = await mistral.chat(mock_model, system_prompt, user_prompt)
+
+    assert response == "An error occurred while processing the request."
+    assert ("src.llm.mistral", logging.ERROR, "Call to mistral api failed: response was None") in caplog.record_tuples
```
gaganahluwalia, line 64: Could you please format this in multiple lines like you did it line 77?

gaganahluwalia (commented):
Looks good, I have made a few comments. Please have a look and see if those make sense.

gaganahluwalia (changes requested):
Please let me know if the comments make sense to you.

backend/src/llm/mistral.py
```diff
@@ -9,21 +8,27 @@
...
+            logger.error("Call to Mistral API failed: No valid response or choices received")
+            return "An error occurred while processing the request."
 
         content = response.choices[0].message.content
+        if not content:
+            logger.error("Call to Mistral API failed: message content is None or Unset")
+            return "An error occurred while processing the request."
 
-        return content if isinstance(content, str) else " ".join(content)
+        logger.debug('{0} response : "{1}"'.format(model, response.choices[0].message.content))
```
gaganahluwalia, line 33: DOes it make sense to just log 'Content' now ?

hsauve-scottlogic (dismissed):
Looks great, thanks for making the variable name change!

gaganahluwalia (approved):
Approved.

hsauve-scottlogic (approved):

